Issue #28, opened 2022-08-10
IMX347 detected as IMX335


ch999dev:
board MC-K45 ( SSC30KQ / IMX347 )

https://github.com/OpenIPC/ipctool/blob/792242207cc594ef38b708143cbd32eb84d3bdbe/src/sensors.c#L91
I think something like this would be better
```
int sensor_id = READ(0x57); // 0x3057
switch(sensor_id) {
  case 0x06: // IMX347
  break;
  case 0x07: // IMX335
  break;
}
```

widgetii:
Please fell free to make a PR, don't have IMX347 to test it but can test on IMX335 still

PR #29, opened 2022-08-10
IMX347 detector

ch999dev:
#28

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -93,11 +93,12 @@ static int detect_sony_sensor(sensor_ctx_t *ctx, int fd,
         sprintf(ctx->sensor_id, "IMX347");
         return true;
     }
+#if 0
     if(chip_id == 0x07) {
         sprintf(ctx->sensor_id, "IMX355");
         return true;
     }
-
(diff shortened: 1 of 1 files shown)


src/sensors.c
```diff
@@ -88,6 +88,17 @@ static int detect_sony_sensor(sensor_ctx_t *ctx, int fd,
     if (i2c_change_addr(fd, i2c_addr) < 0)
         return false;
 
+    int chip_id = READ(0x57);
+    if(chip_id == 0x06) {
+        sprintf(ctx->sensor_id, "IMX347");
+        return true;
+    }
+#if 0
```
widgetii, line 96: I tested both ifdef part on IMX335, and it seems to work. I suggest removing ifdef from the code and stick with one IMX335 detection method (because in a year it will be near impossible to understand why they are both exist)

We can have "backup" method in text comment with detailed explanation in case we need to distinguish more sensors in the future

widgetii:
Thank you for your contribution!

ch999dev:
NP. Thank you for your work!

PR #33, opened 2022-08-18
mips32 correct mac detection, thanks @themactep

cronyx:


src/network.c.old
```diff
@@ -0,0 +1,49 @@
+#include <stdio.h>
```
widgetii, line 1: Please remove this file (was added by mistake?)

cronyx: done

PR #40, opened 2022-09-15
uuencode + script method to upload ipctool

DSchndr:


widgetii:
I guess using copy-paste python script from the documentation and use it is a bad idea (it's not about loosing spaces, but due to this approach is unmaintainable). 

Please consider moving source code into a separate file (say to `tools` directory) and make a link to it from documentation

widgetii:
The idea was implemented in #66

PR #42, opened 2022-09-30
Fix imx290/imx291 detect

p0i5k:


dimerr:
It seems IMX291 detection is broken at the moment. Let's try this approach, thanks

PR #66, opened 2023-04-30
Add alternative telnet upload script

viktorxda:
This commit expands from pull request #40, which i think is an interesting idea.

- Example usage:
```
wget https://github.com/OpenIPC/ipctool/releases/download/latest/ipctool
base64 ipctool > ipctool.b64
tools/base64_upload.py --host 192.168.1.10 --user root --password 12345 ipctool.b64
```

The script doesn't countercheck the telnet credentials, so that needs to be verified prior executing.
Or can be added/extended on a later time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ C library to work on vast majority of hardware.
      46% |**************                 | 61952   0:00:01 ETA
   ```
 
+* *Using telnet/console only*: uses a python script to transfer ipctool via
+  telnet/echo to the camera:
+  ```
+  wget https://github.com/OpenIPC/ipctool/releases/download/latest/ipctool
+  od -An -tx1 ipctool | sed 's/ /\\x/g' > ipctool.hex
+  tools/telnet_upload.py --host 192.168.1.10 --user root --password 12345 ipctool.hex
+  ```
+
 ## Usage
 
 ```console
--- a/tools/telnet_upload.py
+++ b/tools/telnet_upload.py
@@ -1,0 +1,33 @@
+#!/bin/python3
+
+import argparse
+import telnetlib
+
+argparser = argparse.ArgumentParser()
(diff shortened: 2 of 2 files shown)


dimerr:
I'm afraid base64 will be available in 5% of stock firmwares

widgetii:
> * Example usage:

Let's add it into README into `Alternative launch methods` section because nobody remember it next day

themactep:
we have base64 on all platforms and use it in web ui.

dimerr:
Typical list of commands from a XM board
```
ProductDefinition  arping             devmem             gunzip             ifconfig           mkfifo             poweroff           sleep              udhcpc
Sofia              ash                dmesg              gzip               init               mknod              ps                 ssp_read           udhcpd
SofiaRun.sh        awk                dumpleases         halt               insmod             mnt                pwd                ssp_write          umount
Squirrel/          btools             dvrHelper          head               kill               modinfo            reboot             sync               unlink
XmDemux            busybox            echo               hiddrs             killall            mount              rm                 tar                upgrader
XmDvrBox           busybox4           envext             himc               ln                 mv                 rmdir              telnetd            xmmm
XmEnv              cat                false              himd               login              netinit            rmmod              test               xmprop
XmGpio             chmod              flash_eraseall     himd.l             ls                 netstat            route              top
XmServices_Mgr     cp                 free               himm               lsmod              passwd             run_loop           touch
[                  cut                getty              i2c_read           mdev               ping               sed                true
[[                 date               grep               i2c_write          mkdir              ping6              sh                 tty
```

DSchndr:
Wouldn't an option to choose the binary/method (echo, uudecode,...) be better?
On Videopark devices (dahua odm iirc) the shell (or telnet) is inherently broken so you can forget sending the binary like that :)

Probably a cheap mirai patch xD

PR #80, opened 2023-07-28
cYAML: Fix handling of empty objects.

gckzl:
While working on the YAML code, I noticed it didn't handle empty strings, empty objects and empty arrays correctly: in that case, it would output invalid YAML (this limitation was present also before my re-write).  This fixes it.


FlyRouter:
> While working on the YAML code, I noticed it didn't handle empty strings, empty objects and empty arrays correctly: in that case, it would output invalid YAML (this limitation was present also before my re-write). This fixes it.

Thank you very much for the PR, we will look into it shortly.

PR #81, opened 2023-07-29
Sigmastar: add missing standalone library directive

viktorxda:


FlyRouter:
> 

Thank you very much for the PR, we will look into it shortly.

PR #93, opened 2023-09-16
ipcinfo: add info usage label

viktorxda:


FlyRouter (approved):
Thank you very much for the addition of the feature.

PR #105, opened 2023-12-16
sensor: set correct page for superpix detection

viktorxda:


FlyRouter (commented):
Thanks for the additions, I have no objections and I think that @widgetii and @dimerr will look into this soon

src/hal/common.c
```diff
@@ -80,14 +80,15 @@ int universal_i2c_write_register(int fd, unsigned char i2c_addr,
...
                                  unsigned int data, unsigned int data_width) {
     (void)i2c_addr;
-    (void)data;
-    char buf[2];
+    char buf[3];
 
     if (reg_width == 2) {
         buf[0] = (reg_addr >> 8) & 0xff;
         buf[1] = reg_addr & 0xff;
+        buf[2] = data;
```
widgetii, line 88: Can you please describe the purpose of the changes? I don't remember exactly, but it seems, previous version worked on all platforms without it

viktorxda: It sets the data that needs to be written to the i2c device.
I can revert the changes in this case since it only needs to write 0x0, which should be the default value of buf[1].

widgetii: It's clear now

widgetii:
LGTM from my point of view. Anyway sensor detection is tricky, so the only approach is a try and an error method

viktorxda:
I checked the other sensors and 0xFD seems to be a common register: [SP2305](https://github.com/OpenIPC/openxiongmai/blob/main/libraries/libsns_X123_XM530/sp2305_cmos.c#L36).
However, if there are any issues with Superpix in the future this commit can be reverted.

PR #110, opened 2024-01-20
Ingenic: add initial reginfo for T31

viktorxda:


FlyRouter (approved):
Thanks, I think we can enable this to check

PR #114, opened 2024-03-01
include last register in i2cdump output

### LINKED ISSUE #113: i2cdump does not show last register byte
## last byte of requested sequence is missing

`ipctool i2cdump 0x60 0x3107 0x3109` shows only 2 of 3 register bytes `CB 3E`  . The last byte of the register is always skipped:

```
root@openipc-gk7205v210:/tmp# ./ipctool i2cdump 0x60 0x3107 0x3109
       0  1  2  3  4  5  6  7   8  9  A  B  C  D  E  F
CB  3E
```
But output should be (in my case) `CB  3E 01`.
Even though we can get the output we want by incrementing the to-register by 1, the last register will never be readable by i2cdump.

In other words: [source](https://github.com/OpenIPC/firmware/issues/1324#issuecomment-1969332980)

> ipctool is kinda stupid and does not include the last region in the sequence

## [code problem](https://github.com/OpenIPC/ipctool/blob/3e7944bea7b8fe0956cc1f5159cbac33ba4ea850/src/i2cspi.c#L134)
```c
for (size_t i = from_reg_addr; i < to_reg_addr; ++i) {
```

## solution
```c
for (size_t i = from_reg_addr; i <= to_reg_addr; ++i) {
```

RoboSchmied:
`ipctool i2cdump` will now also output the last register (`to_reg_addr`).
closes #113

### DIFF
--- a/src/i2cspi.c
+++ b/src/i2cspi.c
@@ -131,7 +131,7 @@ static void hexdump(read_register_t cb, int fd, unsigned char i2c_addr,
 
     size_t size = to_reg_addr - from_reg_addr;
     printf("       0  1  2  3  4  5  6  7   8  9  A  B  C  D  E  F\n");
-    for (size_t i = from_reg_addr; i < to_reg_addr; ++i) {
+    for (size_t i = from_reg_addr; i <= to_reg_addr; ++i) {
         int res = cb(fd, i2c_addr, i, SELECT_WIDE(i), 1);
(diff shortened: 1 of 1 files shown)


dimerr:
thanks, exactly what was needed

RoboSchmied:
My pleasure, Sir.  :)   Something similar (for i2cdetect) is fixed here #116

PR #125, opened 2024-03-27
refactoring i2cdetect makes it 2 times faster (on goke)

RoboSchmied:
refactoring i2cdetect makes it 2 times faster (on goke)
- open, close and cleanup i2c only once per run

before: `time -p ./ipctool i2cdetect`
```
real 0.14
user 0.02
sys 0.08
```
after: `time -p ./ipctool i2cdetect`
```
real 0.07
user 0.00
sys 0.05
```


dimerr:
great job

PR #128, opened 2024-04-29
Fixes handling of the NOR chip details.

gckzl:
In my initial refactoring of `cb_mtd_info()`, I didn't realize that `nor_chip` contains a preformatted chunk of YAML (I had assumed a string).


widgetii:
Thank you for fixing it!

PR #130, opened 2024-12-26
add 3519DV500

sansarus:
add 3519DV500


dimerr:
Since there are no other ideas, i suggest we take main chip of the family and use it in naming, same as it is with DVR/NVR chips. if Hi3519DV500 is the main, then in macros it'd be HISI_3519D500

PR #137, opened 2025-10-03
Add i2c bus autoselector (just Novatek hal support now)

sansarus:
#define MAX_i2c_NUMBER 5   at sensor.c

### DIFF
--- a/src/sensors.c
+++ b/src/sensors.c
@@ -19,6 +19,8 @@
 #include "sensors.h"
 #include "tools.h"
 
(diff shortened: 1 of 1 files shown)


ystinia (commented):
Hello.
Thank you for your efforts, but the use of variables like the maximum I2C bus pointer is concerning and goes against the utility's intended use. The use of puts methods is also discouraged.
I think we can work on your patch with other developers and then accept it.

dimerr (approved):

PR #179, opened 2026-07-30
sensor: fix IMX415 latching into an IMX335 misdetection via 0x316A

widgetii:
## Problem

A Sony **IMX415** on Hi3516AV300 is reported as `imx335_i2c`. The streamer trusts that answer and brings the sensor up with the IMX335 register table — a 4K sensor forced into a 5MP IMX335 mode.

This is **not** a regression from #167. That PR only touches the `READ(0x3057) == 0x06 → IMX347` branch, and IMX415 reads `0x3057 == 0x00`, so the branch is never entered. #167 is verified still working (see below).

## Root cause

`0x316A` is **INCKSEL4** (input-clock select), not an identifier, and the check
`(r316A & 0xFC) == 0x7C` cannot discriminate. Measured on real silicon:

| state | `0x316A` |
|---|---|
| IMX415, pristine (cold boot) | **`0x00`** |
| IMX415, after `libsns_imx335.so` ran | `0x7E` |
| real IMX335 | `0x7E` |

**It self-latches.** `libsns_imx335.so` *writes* `0x316A = 0x7E` in both its linear and WDR init tables (`sony_imx335/imx335_sensor_ctl.c:238` and `:435`, commented `INCKSEL4 = 37.125`). Streamers ask ipctool which sensor is present, then dlopen that vendor lib. So a single misdetection — from any cause — stamps `0x7E` into an IMX415, and from then on ipctool keeps agreeing with itself. The loop survives streamer restarts and soft reboots; **only removing power clears it**. That is why an affected camera works for months and then "suddenly" breaks and stays broken.

Compounding this, the correct IMX415 test (`0x3B00`) sat *below* the `0x316A` test, so it was unreachable once latched.

## Fix

Rule IMX415 out before concluding IMX335. `0x300B` is the right discriminator: **neither** vendor init table writes it, so it is a true reset default on both parts, and it measured `0xA0` on the IMX415 in **both** the latched and pristine states.

Only reject IMX335 when IMX415 is *positively* identified — the same idiom #167 established for IMX347 — so no sensor loses detection if a register drifts. The looser `0x3B00`-only test is kept below as a fallback.

## Evidence

Full `0x3000`–`0x40FF` dumps (4352 registers) from a real IMX415 and a real IMX335, cross-referenced against the register-address sets each vendor init table writes:

| reg | IMX415 | IMX335 | in imx335 init table? | in imx415 init table? |
|---|---|---|---|---|
| `0x316A` | `0x7E`¹ | `0x7E` | **YES (`0x7E`)** | no |
| `0x3B00` | `0x28`² | `0x00` | no | YES (`0x2E`) |
| **`0x300B`** | **`0xA0`** | **`0x00`** | **no** | **no** |
| `0x30C0` | `0x2A` | `0x00` | no | no |
| `0x341C` | `0x47` | `0xFF` | YES | no |
| `0x3072` / `0x3074` | `0x00` / `0x19` | `0x28` / `0xB0` | YES (WDR only) | no |

¹ latched; pristine `0x00`  ² `0x2E` once the IMX415 driver has run — the code accepts both, which is necessary.

The dropped HINT claiming `0x30C0 == 0x20` on IMX415 is wrong; it measures `0x2A`.

## Verification on hardware

- **Hi3516AV300 + IMX415, still latched** — `imx335_i2c` → **`imx415_i2c`**. The fix works *without* a power cycle, which is the point.
- **Hi3516EV300 + IMX335** — `imx335_i2c`, unchanged. Its `0x3057` reads `0x06`, so it genuinely exercises the #167 fall-through rather than skipping it.
- **Hi3516AV300 after a cold power-cycle** — `imx415_i2c` → `imx415_i2c.ini` + `libsns_imx415.so` → `IMX415 4K2K_10BIT_30FPS init succuss!`, VI pipe **3840x2160** (was 2592x1944 under the wrong driver), IntCnt +80 over 4s = 20 fps, LostFrame 0, VbFail 0.

## Note for future sensor work

Any Sony fingerprint register that appears in a **vendor init table** is a latch hazard, not an identifier — a wrong answer makes the streamer write that table and cement the error. Screening candidates against the init-table address lists leaves 915 registers that separate IMX415 from IMX335 and are written by neither, so there is no shortage of safe choices.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


widgetii:
> **1. IMX415 misdetect on read error** — `is_imx415` is computed using `READ(0xB) == 0xA0` without checking for the read-error sentinel (`-1`).

Valid, and worth fixing — addressed in 46b775c.

Confirmed the sentinel is real rather than theoretical: `hisi_gen1_sensor_read_register` returns `-1` on ioctl failure (`src/hal/hisi/hal_hisi.c:238`), and the code immediately above already respects it (`if (r316A == -1) return false;`). So a failed `0x300B` read really would have read as "not IMX415" and dropped a latched IMX415 into the IMX335 branch — re-arming the exact loop this PR exists to break. That failure mode is unusually expensive here because clearing the latch needs a physical power cycle, not a reboot, so the asymmetry justifies the guard even though the trigger is rare.

I scoped it more narrowly than the suggested fix:

```c
if (r316A > 0 && ((r316A & 0xFC) == 0x7C) && !is_imx415) {
    if (r3B00 == -1 || r300B == -1)
        return false;
    ...
}
```

The guard sits inside the IMX335 branch rather than before it, because only that decision is at risk. Hoisting it would make a failed `0x300B` read abort detection for every Sony part that legitimately falls through past IMX335 (IMX178, IMX274, IMX185, IMX291/307/327, …), which trades a rare wrong answer for a broad new failure mode. The checks below re-test `0x3B00` on their own, so they are unaffected.

Re-verified on hardware after the change: Hi3516AV300 + IMX415 → `imx415_i2c`, Hi3516EV300 + IMX335 → `imx335_i2c`.

Issue #112, opened 2024-02-29
[Same Chip ID] 3 sensors are not distinguished 


RoboSchmied:
## Problem
There are **3 sensors** in the openipc repos that have the **same chip ID  `cb3e`** and are all seen as `sc223a` by ipctool.
- [sc223a](https://github.com/OpenIPC/openingenic/blob/3a9fdfb35c6e45f15d3aaebaf9ca8bd77914c7a1/kernel/sensors/t31/sc223a/sc223a.c#L26) ([30 fps](https://smartsens.oss-cn-beijing.aliyuncs.com/web/products/SC223A_V3.0.pdf))
``` c
#define SC223A_CHIP_ID_H	(0xcb)
#define SC223A_CHIP_ID_L	(0x3e)
```
- [sc233a](https://github.com/OpenIPC/openingenic/blob/3a9fdfb35c6e45f15d3aaebaf9ca8bd77914c7a1/kernel/sensors/t41/sc233a/sc233a.c#L28) ([60 fps](https://smartsens.oss-cn-beijing.aliyuncs.com/web/products/SC233A_V3.0.pdf)) 
``` c
#define SC233A_CHIP_ID_H	(0xcb)
#define SC233A_CHIP_ID_L	(0x3e)
```
- [sc2239p](https://github.com/OpenIPC/openingenic/blob/3a9fdfb35c6e45f15d3aaebaf9ca8bd77914c7a1/kernel/sensors/t31/sc2239p/sc2239p.c#L26)
``` c
#define SC2239P_CHIP_ID_H	(0xcb)
#define SC2239P_CHIP_ID_L	(0x3e)
```

[ipctool handling chip id `cb3e`](https://github.com/OpenIPC/ipctool/blob/3e7944bea7b8fe0956cc1f5159cbac33ba4ea850/src/sensors.c#L593)
```c
    case 0xcb3e:
        // XM
        strcpy(ctx->sensor_id, "SC223A");
        return true;
```
## History

[Here](https://github.com/OpenIPC/firmware/issues/1324) we found out that there may be different "versions" of the sc223a.

Thanks to @johndi3 who [mentioned](https://github.com/OpenIPC/firmware/issues/1324#issuecomment-1970326960) `sc233a` and thanks to @themactep who helped [discovering the ChipID](https://github.com/OpenIPC/firmware/issues/1324#issuecomment-1969336663).

## Solution

[...]

PR #180, opened 2026-07-31
fix SP2305 detection

conke:


widgetii:
Thanks for the patch, but the PR description is empty, so I can't tell what problem this is fixing. Before we change detection behaviour I need the rationale:

- What device / SoC and which sensor module did you observe this on?
- How did you establish the part is an SP2305 and not an OV2735 — silkscreen on the sensor, module BOM, vendor driver name, something else?
- Please post `ipctool i2cdump <sensor-i2c-addr> 0x00 0x0F` (page 0, i.e. after writing `0xFD = 0`) from that device, together with the full `ipctool` output.

The reason I'm asking rather than merging: as far as we can tell SP2305 and OV2735 are the *same silicon* sold under two brands, so `0x2735` doesn't identify one over the other — flipping the label just moves the wrong name onto somebody else's hardware.

- The OmniVision OV2735 datasheet (v2.1, 2017) documents exactly the probe we do here: `P0:0x02 CHIP_ID = 0x27`, `P0:0x03 CHIP_ID = 0x35`, `P0:0x04 ECO_VER = 0x01`, `P0:0xFD` page select. So `0x2735` is OV2735's own documented ID.
- SP2305 vendor drivers use an identical register map — e.g. XM's `sp2305_cmos.c` writes `0xfd=1`, then `0x03`/`0x04` as exposure MSB/LSB, matching the OV2735 datasheet's `P1:0x03 BUF_EXP_8MSB` / `P1:0x04 BUF_EXP_8LSB`. Goke's own `libsns_sp2305.so` even prints `==ominivision sp2305 sensor HCG 1080P30fps(MIPI port) init success!==`. OmniVision and SuperPix have been the same group (WillSemi) since 2018-19, and SP↔OV/OS silicon twins are routine.
- OV2735B is a third marketing name that most likely reports the same ID (separate Ingenic driver and yaml).

There's also a compatibility problem. `ipcinfo --short-sensor` output is consumed programmatically by the firmware load scripts, which `case` on the lowercased name:

- `hisilicon-osdrv-hi3516cv200/.../load_hisilicon` → `ov2710_mipi | ov2735_i2c_1080p | ov2735)`
- `hisilicon-osdrv-hi3516cv300/.../load_hisilicon` → `ov2718_i2c|ov2735_i2c)`
- `ingenic-osdrv-t21/.../load_ingenic` → `"ov2735b")`
- nothing anywhere matches `sp2305`

So as it stands this patch would drop those boards into the `*) UNSUPPORTED sensor` branch — no pinmux, no sensor clock, camera doesn't come up.

If your register dump shows something that genuinely differs between the two labels (`ECO_VER` at `P0:0x04`, or the package register at `P2:0x35`, are the candidates), we can distinguish them properly — the same approach requested in #112 — and I'd be glad to take that patch. If the two are bit-identical, then the correct output is one name plus an alias comment, and the firmware side needs a matching change before the name can flip.

widgetii:
Closing this due to no response

PR #183, opened 2026-08-16
sensors: make getsensoridentity/getsensorshort thread-safe (fix garbage names under concurrent callers)

widgetii:
The sensor-detection entry points are not thread-safe, and a concurrent caller makes them return a **binary-garbage sensor name** (e.g. `getsensoridentity()` yielding `^P\xbf\xbf...` instead of `sc2332_i2c`). Three defects on the path, none of them locked:

1. **`getsensorid()` drives probing through the process-global `i2c_adapter_nr`** and *writes* it in its retry loop (`i2c_adapter_nr = i;`). Two threads probing at once clobber each other's bus number mid-scan, so one probes the wrong adapter and detection fails.
2. **`getsensorid()` falls off the end with no `return`** when every bus fails — the retry `for` loop closes and so does the function, with no `return false`. The `bool` is then indeterminate (often `true`), so the caller treats an all-failed probe as success and reads an **uninitialised `sensor_ctx_t`**.
3. **`getsensoridentity()` and `getsensorshort()` format into one `static char sensor_indentity[16]`** via `lsnprintf` (a `vsnprintf` plus a second in-place `tolower` pass — not atomic). Concurrent calls tear the buffer.

Chained: a concurrent caller → wrong-bus probe (1) → all fail → fall through returning `true` with a garbage `ctx` (2) → that garbage is formatted into the shared buffer, torn by the other thread (3) → a binary-garbage identity string.

Standalone `ipcinfo` never hits this (single process, single thread). It shows up when a program links `libipchw` and calls these from more than one thread — e.g. one thread resolving the sensor at startup while another renders the sensor name on a timer.

## Fix

Serialise detect-and-format under one mutex in the two public entry points (they call `getsensorid()` internally, which does **not** take the lock, so there's no self-deadlock and direct `getsensorid()` callers are unaffected), and give `getsensorid()` the missing `return false` on the exhausted-buses path. With the writes serialised, every returned value is one complete, valid string.

## Verified

Reproduced on an Ingenic T31 (SC2332) by running a second thread calling `getsensoridentity()` in a loop while the main thread resolves the sensor, no `SENSOR` env, 20 starts:

| build | garbage / 20 |
|---|---|
| unpatched | 3 |
| **this patch** | **0** |

`pthread.h` is already included in `src/sensors.c`.


widgetii:
Addressed — `14d3744`.

**Missing pthread link** — fixed. sensors.c's mutex is the first real pthread use in the tree (the other `#include <pthread.h>` are unused, and `fake_symbols.c` even had commented-out pthread stubs), and the targets only linked `m`. Added `find_package(Threads REQUIRED)` and linked `Threads::Threads` on `ipchw` (on the interface, so `ipcinfo` and external consumers of `libipchw` pull it in) and on `ipctool`.

Verified both link paths:
- Standalone host build: `Found Threads: TRUE`, `ipcinfo` links and `nm` shows `pthread_mutex_lock` resolved (`T`).
- As a consumed static lib (a downstream project linking `libipchw`, musl): configures and links clean.

On the approach question your summary raised: agreed the single mutex around detect+format is the minimal-risk fix for the current design. The cleaner long-term shape — drop the global `i2c_adapter_nr` mutation and format into a caller-provided buffer so probes can run truly concurrently — is a larger refactor and worth a separate change; I kept this one small and API-compatible.

PR #187, opened 2026-09-02
cryptobench: per-SoC AEAD seal rate, in software and on the cipher engine

### LINKED ISSUE #186: Proposal: per-SoC AEAD throughput probe (ChaCha20 is 2.3x AES-GCM on GK7202V300)
Proposing a small AEAD throughput probe for `ipctool`, so the per-SoC crypto picture
becomes data instead of an assumption.

@widgetii suggested this home in OpenIPC/firmware#2344 — worth reading that thread for
the full context, but the short version follows.

## Why

Hardware crypto acceleration is currently enabled for two platforms because someone
measured those two platforms. Every other SoC OpenIPC ships is an inference. And the
measurements that exist show the answer is genuinely per-chip, not per-family:

- On a **gen-3** part, the crypto engine measured *slower* than software, so it is not
  built there.
- On **gen-4 HiSilicon**, the engine is a bit over 2x software (86.4 → 38.4 µs and
  87.6 → 42.0 µs per 1100-byte packet).
- On **GK7202V300** (Goke, ARMv7 Cortex-A7, no ARMv8 crypto extensions, no engine
  OpenIPC builds for), I measured in software with `ring` 0.17 — the same backend the
  Rust TLS/QUIC stacks use — sealing 1200-byte packets:

  | AEAD | throughput | per packet |
  |---|---|---|
  | AES-128-GCM | 72.8 Mbit/s | 131.9 µs |
  | AES-256-GCM | 58.7 Mbit/s | 163.4 µs |
  | **ChaCha20-Poly1305** | **166.0 Mbit/s** | **57.8 µs** |

  ChaCha20 is **2.3× AES-128-GCM** here. NEON carries ChaCha20; AES has no hardware to
  stand on. That inverts the usual server-side ordering where AES-NI makes AES-GCM the
  obvious default.

That my software AES-128 figure (131.9 µs/1200 B) lands near the HiSilicon *software*
numbers on different silicon is a decent cross-check that software AES on this
generation is simply expensive — and that the interesting variable is what each chip
offers to accelerate it.

A table with a row per SoC turns "enable the engine here?" into a lookup.

## What I am proposing

A probe that reports, for the chip it is running on:

- AES-128-GCM, AES-256-GCM and ChaCha20-Poly1305 seal throughput
- measured on **packet-sized buffers (~1200 B), not bulk**

The buffer size matters and is the one design point I would not compromise on. Bulk
throughput flatters AES on chips where per-call setup dominates, and every real consumer
here — TLS records, QUIC packets, SRTP — is packet-sized. Bulk numbers would produce a
table that looks authoritative and mispredicts the thing we care about.

Optionally, where a crypto engine exists, the same measurement through it, so the
engine-vs-software delta is directly comparable rather than inferred.

## Practicalities I would want input on

1. **Language/deps.** My implementation is ~40 lines of Rust against `ring`, which is
   not what `ipctool` is written in. If this belongs in-tree it presumably wants C. I am
   happy to port it — but if the point is to characterise what *our software stacks*
   actually achieve, then measuring the library those stacks use has some value that a
   generic C implementation would lose. I do not have a strong view; you know the tree.

2. **Where the numbers go.** The wiki was suggested for the resulting table. Happy to
   seed it with the GK7202V300 row and whatever else people run.

3. **Scope.** Is this the sort of thing `ipctool` should carry at all, or is it better
   as a standalone one-shot binary people run once per board? It is a capability probe,
   which felt like `ipctool`'s remit, but I would rather ask than assume.

I am not an `ipctool` contributor and have been around OpenIPC about three weeks, so if
the answer is "interesting, wrong repo" or "we would rather not grow the tool this way",
that is a perfectly good outcome — say so and I will just publish the numbers on the
wiki instead.

widgetii:
Closes #186.

Adds `ipctool cryptobench`, beside `cpubench` and `membw`, which are the same kind of per-SoC probe:

```
ipctool cryptobench [--json] [--bytes N] [--iters N] [--no-hw]
```

## The finding that reframes the proposal

The proposal offers the hardware engine as the thing that might rescue AES-GCM on these parts. It cannot, and not because it is slow — **gen-4 HiSilicon and Goke silicon has no AEAD mode at all.** `CHIP_AES_CCM_GCM_SUPPORT` is defined for `hi3569v100` alone; the block does ECB/CBC/CTR/CFB/OFB and nothing else. There is no hardware sealing to enable, on any part OpenIPC currently ships.

`cryptobench` establishes that by *asking* — it configures a channel for AES-GCM and reports what the driver answers — rather than by asserting it, so a part that does carry the mode reports itself instead of inheriting a hardcoded "unsupported".

What the engine can do is AES-CTR, the confidentiality half of AES-GCM. The authentication half would still be GHASH on the CPU, which is the expensive half. So ChaCha20's advantage on these cameras is not a software artifact waiting to be engineered away. It is the answer.

## Numbers

Lab gk7205v200 (Cortex-A7), 1100-byte packets, majestic stopped, µs per packet:

| | wall | cpu | MB/s |
|---|---|---|---|
| AES-128-GCM | 141.4 | 141.4 | 7.4 |
| AES-256-GCM | 167.6 | 167.6 | 6.3 |
| **ChaCha20-Poly1305** | **45.1** | **45.1** | **23.3** |
| engine AES-CTR, one packet per ioctl | 83.9 | 44.1 | 12.5 |
| engine AES-CTR, 15 packets per ioctl | 49.7 | 12.3 | 21.1 |
| engine, 16 bytes — the per-ioctl floor | 37.4 | 34.1 | — |

**ChaCha20 wins by 3.14×**, so your reported ordering holds and is wider on this part. Different chip and a different implementation from `ring`, so I would not read much into 3.14 vs 2.3 beyond the direction — which is the argument for having the probe in-tree and reproducible rather than comparing two people's one-off harnesses.

The engine rows carry their own lesson, which is why the 16-byte row is there: **nearly half the cost of a 1100-byte packet is the trip, not the cipher.** A burst of 15 packets under one ioctl (OpenIPC/openhisilicon#217) cuts CPU per packet from 44 µs to 12 while wall clock barely moves — the engine is a poor bargain per packet and a good one per burst.

## Your three questions

**1. Language and deps — C, in-tree.** Not a style preference: `ipctool` ships as a static musl binary (`CMakeLists.txt:9-12`, and `release.yml` builds with the default), musl's `dlopen` is a no-op when static, and all three CI toolchains carry only `libmbedcrypto.so` with no `.a`. The platform mbedTLS is genuinely unreachable from the shipped binary, so linking it would trade the single static binary for a benchmark.

That does cost the thing you correctly flagged — measuring the library our stacks use — so the implementation is built to keep the comparison honest rather than to be small: four runtime-generated AES T-tables and 4-bit GHASH tables, which is what mbedTLS itself does by default (`MBEDTLS_AES_ROM_TABLES` is off). A compact S-box implementation would be several times slower and would inflate ChaCha20's lead into a number about this file rather than about the silicon.

**2. Where the numbers go — the tool generates the table.** Output is YAML by default and `--json` on request, so the wiki holds pasted tool output rather than hand-typed numbers, and a row can be regenerated by whoever doubts it.

**3. Scope — yes, it belongs here.** It is a capability probe of the running chip, which is exactly `cpubench` and `membw`'s remit, and it reads `/dev/cipher`, so it needs to run on the camera anyway. Cost is 19–26 KB of `.text` (arm32 +19252, arm64 +22248, mips32 +26128); all three CI targets cross-build clean and stay statically linked.

And for what it is worth: "interesting, wrong repo" was not the right answer. The proposal was right that this is per-chip and right that packet size is the design point not to compromise on — `--bytes` defaults to 1200 for that reason.

## Correctness before numbers

A benchmark of a wrong implementation is worth nothing, and a wrong one is invisible from the outside because ciphertext is supposed to look like noise. Every primitive is checked against a published vector before it is timed, and a row that fails is reported `verified: false` with no numbers rather than as a fast wrong answer:

- AES-128/256-GCM — GCM spec test cases 2 and 14
- ChaCha20-Poly1305 — RFC 8439 §2.8.2
- engine AES-CTR — NIST SP 800-38A F.5.1, **plus a vector spanning a 32-bit counter wrap**
- engine AES-CTR batched — fifteen packages, **a distinct IV and distinct plaintext each**, every one compared against the software CTR above

The batch vector is separate from the single-packet one on purpose, and it earns its keep: a driver that accepted the burst but applied the channel's one IV to all fifteen packages would return correct ciphertext for package 0 and silent nonsense for the rest, which is precisely the shape of the benchmark's real jobs. It passes on a gk7205v200, so OpenIPC/openhisilicon#217 does honour per-package IVs — previously assumed, now checked. It also runs at a length that is not a multiple of the block, which establishes that the driver rounds the descriptor itself.

The counter-wrap vector is not optional either. An engine carrying only the low 32 bits of the counter agrees with software for every IV that is not near a wrap, which is nearly all of them — it would pass F.5.1, ship, and then disagree on one packet in millions with nothing to say so.

The §2.8.2 vector earned its keep during development: the first cut of this used Poly1305's own short-final-block padding instead of RFC 8439's `pad16()` for the MAC input. That is correct for any message which is a multiple of 16 bytes and wrong for every other one, and nothing but the vector would have caught it.

## Measurement discipline

Three things the numbers depend on:

- **Both clocks on every row.** The hardware rows differ by ~2× between them, because the driver sleeps on its completion interrupt and hands the core back — that gap *is* the offload. Quoting either clock alone hides the point. `cpu_wall` near 1.00 on a hardware row means the run was contending for CPU; `--help` says to stop majestic.
- **Nothing is printed between timed blocks.** Results are buffered and printed once. A single `printf` to a serial console or an ssh pipe between two identical runs has been measured making the second one read twice as slow.
- **Packet-sized, and the floor is reported.** The 16-byte row is almost entirely syscall and channel programming, and it is the number to reason with before building anything on this engine.

## Degradation, tested on the board rather than written

- no `/dev/cipher` (`rmmod open_cipher`) → `hardware: absent`, software half still runs
- a driver that rejects the batched command → `batched: unsupported`, single-packet and floor rows still measured
- non-HiSilicon hardware → same absent path

The `/dev/cipher` ABI is transcribed rather than linked, the way majestic does it: `libhi_cipher.so` is a thin ioctl wrapper and most images do not ship it, so a `DT_NEEDED` on it would stop `ipctool` starting on the very cameras it is meant to inspect. Commands are derived from struct sizes with the vendor's own macro, so a struct that does not match the driver's yields a command it does not recognise rather than one it misreads.

## Second commit, unrelated but found by this work

`cYAML_Print()` returned `NULL` — meaning `ipctool` printed **nothing at all** and exited 0 — if any string anywhere in the document contained a non-ASCII byte. `char` is signed, so every UTF-8 byte tested as a control character, fell into the `\u` expansion and overran its scratch buffer. JSON mode was unaffected, so the two output modes disagreed about whether the machine had any hardware. Sensor names and U-Boot environments are passed through verbatim, so this was reachable from ordinary output; my `note:` strings are just what tripped it. Fixed by comparing as `unsigned char`, with two test cases. The test harness was also computing a per-case verdict and then unconditionally returning 0, so it now exits non-zero when a case fails.


widgetii:
Thanks — finding 4 was a real defect and a well-aimed one. Dispositions below; fixes in ecca9ce.

**4. Batch IV handling unverified — fixed, and it was worse than described.** The probe's own comment claimed it used "two packets rather than one, because one would also succeed on a driver that ignored the per-package IV entirely", and then submitted two packets with the *same* zero IV and never looked at the output. So the comment described a check that did not exist, and the row it enabled inherited `verified` from the single-packet vector on top of that.

There is now a known-answer test for the batched command specifically: fifteen packages, each with a distinct IV and distinct plaintext, every one compared against the software AES-CTR in this same binary (which F.5.1 has already pinned to the standard). `batched.verified` stands on that vector instead of inheriting, and the row is not measured unless it passes. All fifteen match on a gk7205v200, so OpenIPC/openhisilicon#217 does honour per-package IVs — previously assumed, now checked.

**6. Partial batch packets — answered by that same test, no padding added.** The new vector runs at the benchmark's own length, which on that board is 1100 bytes by default and is *not* a multiple of the block. It passes, so the driver rounds the descriptor itself and the two paths agree. Padding in the batch path would need a bounce buffer per package and would make the benchmark measure that copy rather than the engine, so the asymmetry stays — now documented as deliberate and verified, rather than left looking like an oversight.

**2. SoC identity — fixed.** Correct, and it undercut the point of the PR: #186 wants a table with a row per SoC, and a row that does not say which SoC is not a row. Now tagged with `getchipname()` following `membw`.

**7. Iterations can wrap — fixed.** `--iters -1` became `UINT_MAX` and `--iters 100junk` silently became 100. Both options now require the whole argument to parse and are range-checked before narrowing.

**3. Supported GCM remains unmeasured — reported honestly rather than implemented.** No part that accepts AES-GCM exists to test against: `CHIP_AES_CCM_GCM_SUPPORT` is defined for `hi3569v100` alone, and nothing OpenIPC ships is that part. Writing a sealing path that has never executed and cannot be checked against a vector would produce exactly the failure mode this tool is built to avoid — a fast number nobody can trust. That path now reports `aead: supported, not measured` and asks for an issue to be opened, which is a lead to hardware rather than a guess about it.

**5. Invalid UTF-8 emitted raw — fixed properly.** Fair: my change traded a total-output failure for a possibly-unparsable one, which is better but not right. cYAML now validates sequences instead of assuming every byte ≥ 128 is well-formed. Valid sequences pass through unescaped; invalid bytes, overlong forms, surrogates and out-of-range code points become `\u00XX`, which round-trips losslessly and keeps the document parsable. Test added.

**1. `--bytes` permits non-packet AEAD — declining.** Choosing the size is the feature, not a gap. The default *is* 1200, every run states its own `packet_bytes`, and the 16-byte row exists specifically to expose how much of the cost is per-call rather than per-byte — that row is only obtainable because the size is selectable. A benchmark that refused to measure what it was asked to measure would answer a narrower question than #186 poses, and would not have produced the finding that half the engine's per-packet cost is the trip.

PR #190, opened 2026-09-14
pinmux: publish the pad tables through libipchw

openipc-ai:
`src/reginfo.c` holds 1333 hand-entered pad rows across sixteen HiSilicon
tables plus SigmaStar and Ingenic: for each pad, the physical register, the
selector value that puts each named function on it, and the GPIO the pad
carries otherwise. All of it was executable-only, so the only way to ask
"which register and which value select PWM0 on this chip?" was to run
`reginfo` and parse the output — which is why consumers hand-copy the rows
they need, and why those copies go stale.

This publishes the data through `libipchw` without moving a single table row,
so `git blame` still answers for every address.

### Commits

1. **`gpio mux` writes the selector field, not the low half-word.** The write
   was `val & 0xfff0 | new_func` against a 32-bit register, which zeroes bits
   16–31 along with the selector. The selector width is not a constant either
   — low nibble on HiSilicon/Goke, low half-word on SigmaStar — and
   `dump_regs()` knew that by matching vendor strings while
   `fill_enabled_gpios()` had separately open-coded a bare `0xf`. All three now
   share one `padmux_func_mask()`.

   In fairness: I read every pad register on four lab boards (hi3516cv100,
   hi3518ev200, hi3516cv300, hi3516av300) and none of them currently has a bit
   above 15 set, so this is a correctness fix with no victim I can point at,
   not a bug report. Keying the mask on `chip_generation` instead of the vendor
   string is what makes the lookups below testable off-camera.

2. **`reginfo.c` becomes linkable.** `regs_by_chip()` returned
   `exit(EXIT_FAILURE)` for an unknown SoC — inside a long-lived daemon that is
   an outage rather than a refusal, so it returns NULL and the CLI keeps its
   old behaviour through a `regs_by_chip_or_die()` wrapper.
   `gpio_possible_ircut()` takes the softer path instead and drops the ircut
   hint from its report rather than killing the process that asked for it. One
   `#ifndef STANDALONE_LIBRARY` fence after `regs_by_chip()` takes the
   `print_usage()` link edge and every `printf` with it. `get_function()` had
   no callers and is gone.

3. **The lookups.** `ipchw_padmux_by_func`, `_by_prefix` and `_by_pad` on
   `include/ipchw.h`, filling a flat POD so `muxctrl_reg_t` stays private.
   Each returns the number of *matches* rather than the number written, so a
   caller whose array was too small learns the size it needed.

   Two things about this data surprise people, so both are in the contract and
   both have a test:

   - **A function is not unique to a pad.** hi3516ev300 offers PWM2 and PWM3 on
     three pads each and hi3516cv200 offers PWM0 on two, so "the pad for PWM3"
     is not a question with an answer. Callers enumerate.
   - **The spelling is not portable.** `PWM_OUT0` on the V1 parts, `PWM0` from
     V2 on, `PWM0_OUT1` on V5 — which is what the prefix query is for. But
     `SVB_PWM` and `PMC_PWM` are *different controllers*, so the prefix match
     is anchored at offset 0, never a substring search.

   ev200 and ev300 share one SDK build and one register map but not one pad
   table, so the family cannot be resolved at compile time; the lookups read
   `chip_generation` and `chip_name`.

4. **`IPCHW_PADMUX`, so a consumer pays for the SoCs it targets.** The tables
   are 46 KB on arm32 and `regs_by_chip()` names every one of them from a
   single switch, so `--gc-sections` drops nothing. Same shape as
   `IPCHW_VENDORS` directly above it. Measured, `-Os -DSTANDALONE_LIBRARY`,
   `reginfo.c` alone:

   | selection | bytes |
   |---|---|
   | all ten families | 46207 |
   | `v4` (ev200/ev300/gk7205/dv200 — four tables) | 13613 |
   | `v4a` | 9811 |
   | `v2` | 5893 |
   | `v1` | 4667 |
   | `v3` | 4330 |
   | `none` | 353 |

   The macros are PUBLIC on the `ipchw` target so a consumer that trims the
   wrong family gets a compile error rather than `IPCHW_PADMUX_NO_TABLE` on a
   camera. The SigmaStar and Ingenic tables now follow their existing
   `IPCHW_VENDOR_*` macros — 10.6 KB nobody building for HiSilicon was using.

5. **CI runs the unit tests.** `cYAML_test` has existed for years and no
   workflow ever ran it. The new native job builds and runs it alongside
   `reginfo_test`, then configures a trimmed build and runs the tests again.
   Also writes down what `mem_reg()` does not promise: it is not thread-safe
   (one cached window in four file-statics, no lock), and a `false` return
   means "/dev/mem would not open or mmap", not "that address is bad" — a bad
   address on live silicon raises SIGBUS.

### Verification

- `reginfo_test` sets `chip_generation`/`chip_name` by hand and asserts rows a
  real camera was measured against, so an edit that moves PWM1 off
  `0x100C0010` fails there rather than on a bench. It links the real
  `libipchw`, which is what proves the lookups are exported and not merely
  present. It also sweeps every compiled-in family for two invariants worth
  having over 1333 hand-entered rows: no selector wider than its own field, and
  no GPIO name that fails to parse. Both hold today across all sixteen tables.
- Cross-built with the OpenIPC hi3516cv100 toolchain and run on a lab
  hi3516cv100: `reginfo` output is **byte-identical** to the stock binary's, 87
  rows, and `gpio scan` still works. Same check on a lab hi3516ev300 (93 rows,
  PWM0–PWM3 decoded on the pads they belong to).
- `-Wall -Wextra` clean on arm32 across `v1`, `v4`, every family and `none`.
- The `ipctool` executable is byte-for-byte the size it was.

### Note for reviewers

`CV500regs` genuinely lacks the HDMI and power-sequencer pads
(`0x114F0000`–`0x114F000C`, `0x112F00B0`, `0x112F00B8`) that `DV300regs` has,
and that is **correct** — hi3516cv500 has no HDMI. I checked, because a
downstream consumer carries one merged table for all three V4A chips and I
assumed the omission was a gap. It is the merged table that is wrong.


openipc-ai:
Good catch, and it was worse than it looked. Fixed in 3ee8992.

`fill_enabled_gpios()` got its `return false` for "no pad table" and never got the `return true` for the ordinary path, so on **every supported SoC** both callers were deciding whether the platform is usable from whatever happened to be in the return register. It passed CI, and it passed a run on a lab hi3516cv100, because the garbage was truthy.

What let it through is that the default flags carry `-Wextra` but not `-Wall`, and `-Wreturn-type` lives in `-Wall`. So the fix is the flag as well as the line — `-Werror=return-type` is now on in every configuration, and it immediately found four more of the same shape that were already in the tree:

- **`backup_with_yaml()` in `main.c`** computes `do_backup()`'s verdict into `ret` and then drops it, so `ipctool backup <file>` has been exiting on whatever was in the register. That one is user-visible.
- **`dummy_change_addr()` in `sensors.c`** is an empty `static int` stub wired into the `i2c_change_addr` pointer; callers that check it were reading rubbish.
- **`cv100_sensor_clksel()`** and the RAW_DATA name lookup in **`ptrace.c`** are switches with no default. The first is genuinely exhaustive over a 3-bit field and now says so; the second decodes a value that came off a traced ioctl and is not bound to the enum at all, so "unknown" is a real answer it needed.

Verification, since the original also survived a smoke test:

- lab hi3516cv100 — `reginfo` and the full YAML report are both **byte-identical** to the stock binary's output, and the executable is the same size.
- lab hi3516ev300 — this board actually exercises the path the bug was on, and it reports `possible-IR-cut-GPIO: 10,11`. That is `fill_enabled_gpios()` returning true on the success path, on real silicon.

PR #192, opened 2026-09-15
padmux: make ipchw_padmux_* answer on SigmaStar and Ingenic

### LINKED ISSUE #191: padmux: SigmaStar and Ingenic tables carry no pad-mux data, so ipchw_padmux_* answers nothing on either
## Summary

`ipchw_padmux_*` answers usefully on HiSilicon and Goke and answers nothing
useful on SigmaStar or Ingenic. The header already admits this:

> Only HiSilicon and Goke parts carry real alternatives per pad. The SigmaStar
> and Ingenic tables name one function per register, so on those `gpio_name`
> is NULL, `gpio_pad` is -1, and a lookup by pad finds nothing.

That sentence describes the data, not a property of the silicon: both families
do have per-pad function selection. The result is that anything built on this
API — "what is pad N carrying right now", "put pad N back to GPIO", "which pads
can carry I2C1" — works on one vendor and is simply absent on the others.

Worth stating up front, because it decides how much of this is code and how
much is data: **the GPIO side of the API is already derived generically.**
`row_gpio()` in `src/reginfo.c` scans a row's `funcs[]` for the first entry
that parses as `GPIO<bank>_<pin>` and takes its *index* as the selector value
that restores GPIO. Nothing in `padmux_walk()` is vendor-specific. A table
whose rows are shaped like the HiSilicon ones gets the whole API for free.

The two platforms are broken in different ways, and only one of them is a data
fill.

---

## SigmaStar — a data problem

`src/hal/sstar_reginfo.h` has the right register addresses and the wrong
contents. Every row names **the pad** rather than the functions selectable on
it:

```c
MUXCTRL(I6B_reg0,  0x1F207800, "PAD_GPIO0")
MUXCTRL(I6B_reg14, 0x1F207850, "PAD_FUART_RX")
MUXCTRL(I6B_reg18, 0x1F207980, "PAD_I2C0_SCL")
```

Compare a HiSilicon row, where `funcs[]` is indexed by the selector value:

```c
MUXCTRL(EV300_iocfg_reg92, 0x120C0020, "JTAG_TDI", "SPI1_CSN1", "GPIO2_0",
        "SDIO0_CARD_POWER_EN_N", "PWM3", "I2S_SD_TX")
```

So on SigmaStar every row reports `func = 0` whatever the register actually
holds, no row has a GPIO alternative, and `ipchw_padmux_by_pad()` returns
nothing for every pad.

**What it should look like.** One register per pad is already the right model
for SigmaStar; only the payload changes. Each `funcs[]` becomes the list of
modes that pad's config register selects, indexed by the register value, with
the GPIO mode spelled `GPIO<bank>_<pin>` so `row_gpio()` finds it:

```c
/* PAD_GPIO0 on Infinity6B: mode 0 is GPIO, 1..n are the peripherals. */
MUXCTRL(I6B_reg0, 0x1F207800, "GPIO0_0", "UART1_RX", "PWM0", ...)
```

Three things to get right, and they are the whole job:

1. **Index == register value.** A mode that does not exist at some value must
   be `"reserved"`, which `padmux_walk()` already skips, rather than the list
   being compacted — otherwise every later mode is off by one.
2. **The GPIO name must be the running pad number**, `bank*8+pin`, because that
   is what `gpio_name_to_pad()` parses and what every consumer speaks. If
   SigmaStar's own numbering does not divide into banks of eight that way, say
   so on the issue before filling the table — it changes what the mapping has
   to be.
3. **`padmux_func_mask()` returns `0xffff` for the Infinity6 family.** If the
   mode field is narrower (4 bits is typical), fix the mask at the same time,
   or a row will match on bits that are not the selector.

Tables to fill: `I6B_regs` (97 rows), `I6C_regs`, `I6E_regs`. Source is the
vendor's pad-mux documentation or the `padmux` node in the vendor kernel's
device tree, which lists the modes per pad in selector order.

---

## Ingenic — a model problem, not a data one

`src/hal/ingenic_reginfo.h` is not a pad-mux table at all. These are the GPIO
port control registers:

```c
MUXCTRL(T31_reg0,  0x10010000, "PAPIN")
MUXCTRL(T31_reg1,  0x10010010, "PAINT")
MUXCTRL(T31_reg4,  0x10010020, "PAMSK")
MUXCTRL(T31_reg7,  0x10010030, "PAPAT1")
MUXCTRL(T31_reg10, 0x10010040, "PAPAT0")
```

`PAPIN`, `PAINT`, `PAMSK`, `PAPAT0/1`, `PAFLG` are port A's pin state,
interrupt, mask, pattern and flag registers. They are useful for `reginfo`
output and they are not what `ipchw_padmux_*` is for.

Ingenic selects a pin's function with **two bits spread across two registers**:
`INT`/`MSK` choose device-versus-GPIO, and `PAT1:PAT0` choose which of four
device functions — one bit each, at the pin's own bit position, in two
different registers. That cannot be expressed as
`{one address, one contiguous mask, one value}`, which is what both
`muxctrl_reg_t` and the public `ipchw_padmux_t` assume.

So Ingenic cannot be fixed by filling a table, and I would not encourage
bending `ipchw_padmux_t` into a shape that carries a second address and a bit
index — that leaks one vendor's register layout into every consumer of the API,
including the ones that only ever wanted to ask a question.

### Suggested direction: move the access inside ipctool

The API currently hands out a physical address and a mask and leaves every
caller to do its own read-modify-write through its own `/dev/mem` window. That
is what makes Ingenic inexpressible, and it is also a hazard on the platforms
that do work — the header already warns that `mem_reg()` keeps one
unsynchronised mapping and that every write must come from a single thread.

Two accessors would close both:

```c
/* What pad `pad` is carrying right now. Returns the row, or IPCHW_PADMUX_*. */
int ipchw_padmux_get(int pad, ipchw_padmux_t *out);

/* Put `func_name` on pad `pad`. */
int ipchw_padmux_set(int pad, const char *func_name);
```

With those, a consumer never composes a register write, and each vendor's
mechanism stays behind the library — a selector field on HiSilicon and
SigmaStar, an INT/MSK/PAT1/PAT0 quartet on Ingenic. The existing lookups can
stay exactly as they are for the "what could this pad be" question; only the
"what is it now / make it so" half needs the new shape.

If that is too large a change to take in one go, the useful order is:
**SigmaStar tables first** (self-contained, no API change, unblocks that
family), then the accessors, then Ingenic behind them.

---

## Why it matters

OpenIPC's WebUI has a pin page that draws the chip, says what each pad can be
and what it is carrying, and lets an owner assign one — and a scan that hunts
the IR-cut wiring by driving pads, which now refuses to drive a pad the chip
says is currently the serial console. Both read this API. On a SigmaStar or
Ingenic camera the pin page is blank and the scan drives blind, purely because
the table is empty.

That is the argument for the table living here rather than in each consumer:
one place that knows what a pad is, and every tool that asks gets the same
answer.

## How to check the result

`ipctool reginfo` on a board of each family should list the pads with their
current function resolved, and for any GPIO-capable pad
`ipchw_padmux_by_pad(N, ...)` should return more than zero rows with
`gpio_pad == N`. Today both are empty on SigmaStar and Ingenic and populated on
HiSilicon, which is the difference to close.

widgetii:
Closes #191.

`ipchw_padmux_*` answered usefully on HiSilicon and Goke and answered nothing
on SigmaStar or Ingenic. The header admitted it, and the issue took that
sentence at face value: SigmaStar a data fill, Ingenic a model change.

**Half of that is wrong, and it decides the shape of the fix.** Both existing
tables were never pad-mux tables at all. Both were added as "initial gpio
information" (cb4eaec, 568f0c9) and both are the GPIO *controller* registers.
`mhal_gpio.c` settles it for SigmaStar: `GPIO0_OEN` is `0x103C00` bit 5, and
`0x1F000000 + (0x103C00 << 1)` is `0x1F207800`, which is `I6B_reg0`.

## Three vendors, three mechanisms

| | how a pad's function is selected |
|---|---|
| HiSilicon/Goke | a selector field in the **pad's own** register, indexing `funcs[]` |
| SigmaStar | a field per **peripheral**, whose value picks which **group of pads** carries it |
| Ingenic | **four bits in four registers** at the pin's own position, no field at all |

`reg_fuart_mode` on Infinity6B0 is three bits of one chip-top register: value 1
puts FUART on the four `PAD_FUART_*` pads, 2 on `PAD_GPIO0..3`, 4 on
`PAD_SD1_IO0..3`. So one pad's alternatives are fields in *different*
registers, the groups overlap, and a pad is plain GPIO when nothing claims it.
There is no list to index, and SigmaStar is a model problem too.

So this adds a `padmux_ops_t` seam rather than a second table format. **Not one
HiSilicon row moved** — the rows every chip yields are byte-identical before
and after, checked by dumping `ipchw_padmux_by_prefix("")` for all 22 chip
cases and diffing.

| | before | after |
|---|---|---|
| SSC325 / SSC333/335/337 | 0 rows | 589 |
| SSC338Q / SSC30KQ | 0 | 1015 |
| SSC377 | 0 | 721 |
| T31 | 0 | 241 |

## The API gains the half that was missing

```c
int ipchw_padmux_get(int pad, ipchw_padmux_t *out);
int ipchw_padmux_set(int pad, const char *func_name);
```

The lookups hand out an address and a mask and leave every caller to compose
its own write. That is what makes Ingenic inexpressible, and it is a hazard on
the family where it works, because `mem_reg()` keeps one unsynchronised window
the caller now shares. With these, a consumer never composes a register write.

`ipchw_padmux_t` gains `flags`, because a caller composing its own write now
has to know whether that is the whole operation: `F_RMW`, `F_SHARED_REG`,
`F_GPIO`. `func_mask` is documented as the field **in place** rather than
based at bit 0 — a documentation fix for HiSilicon, load-bearing for
SigmaStar, and no behaviour change for anyone (the value was always applied
unshifted).

`IPCHW_PADMUX_ADDR_NONE` is `0xdeadbeef` rather than 0 on purpose: `mem_reg()`
already treats that address as a no-op, while 0 is its *teardown* call. A
consumer that ignores `flags` and composes `(old & ~0) | func` from a
non-writable row therefore writes the register back unchanged, instead of
unmapping the window under everything else.

## The tables are generated

`tools/gen_sstar_padmux.py` and `tools/gen_ingenic_padmux.py`, both with
`--verify`. 2900 rows is past the point where transcription is honest, and the
generator can check what a reader cannot: that mode → (register, mask, value)
is a function, 135/135 modes on i6b0, which is what lets a pad hold mode
indices instead of triples. They refuse rather than guess — a value with bits
outside its field, a pad `gpio.h` does not name, a mode with two register
tuples, a row under the wrong port's table.

CI has no SDK and does not regenerate. What stands in for it is `reginfo_test`,
which installs a fabricated register file through a `padmux_io_t` seam, sweeps
five structural invariants over every compiled-in family, and then sets every
function of every pad and reads it back — ~3800 rows. Breaking one of Ingenic's
four writes, or the `PZGID2LD` commit, fails 84 checks.

## Two judgement calls worth a reviewer's eye

- **SigmaStar rows carry `F_RMW` but not `F_SHARED_REG`, and `gpio_func` is -1
  on all of them.** Getting back to GPIO means dropping every claim, which is
  not one write. majestic's `mux_gpio()` already refuses on `gpio_func < 0`, so
  it degrades to a refusal rather than a wrong register write, with no change
  in that repo.
- **Twelve i6b0 pads answer nothing rather than "GPIO"** — PM_GPIO4, the SAR,
  ETH and USB pads, which the vendor programs from a hand-written switch
  wanting several banks and a `0xBABE` unlock. They were reading as free GPIO
  because on SigmaStar free GPIO is the absence of every claim. A pin page
  offering the Ethernet pair as a wire to drive is worse than one missing four
  pads.

## What `reginfo` does now

Unchanged on every family: the SigmaStar and Ingenic register dumps are useful,
they were simply never pin-mux. `reginfo --pads` is the new per-pad view, and
is the issue's own acceptance check. `gpio mux` routes through
`ipchw_padmux_set()`, so it means the same thing on a vendor where the write is
not one register; the numeric form still works by resolving a selector value to
the name of the function it selects.

## Verification

Host: native, arm32, mips32 and arm64 build clean; six knob configurations
including `IPCHW_PADMUX=none` and `IPCHW_VENDORS=none`; `cYAML_test`,
`reginfo_test`, `test_pipeline.sh`, both `--verify`. arm32 `ipctool`
300320 → 328992 bytes (+28 KB for four SigmaStar families and T31).

**Ingenic, on hardware** — T31L "Swan", sc2332, kernel 3.10.14, majestic
running throughout:

- `reginfo` and the full YAML report byte-identical to the base-commit binary,
  except `PAPIN`/`PBPIN`, which are the live pin-input registers and change
  run-to-run on the stock binary too.
- `reginfo --pads` lists 77 pads; decoding the live `PxINT`/`PxMSK`/`PxPAT1`/
  `PxPAT0` independently agrees on **77/77**.
- Five corroborations from the running system that the table itself is right:
  `console=ttyS1` ↔ PB23/PB24 = UART1_TXD/RXD; `nogmac` with only `wlan0` and
  two mmc hosts ↔ the GMAC pads muxed to MSC1; `mtdparts=jz_sfc` ↔ PA23/24/27/28
  = SFC, with PA25/26 left GPIO because a single-chip NOR uses CE0 only;
  `i2c-0` device `0-0030` ↔ PA12/PA13 = SMB0_SDA/SCK; DVP sensor ↔ PA15 =
  DVP_MCLK. PB19/PB22 sit at TDI/TDO, the reset default the GPIO spec's own
  note predicts for those pins.
- Write path on PC15, a GPIO input whose alternatives are unused on that board:
  `[GPIO]` → `gpio mux 79 UART0_CTS` → `[UART0_CTS]` → `gpio mux 79` → `[GPIO]`.
  `PCMSK`/`PCPAT1` `0x07fffbff` → `0x07ff7bff`, `PCPAT0` `0` → `0x8000`,
  `PCINT` untouched — exactly bit 15 in exactly three registers, the other 31
  pins of the port left alone, reverted byte-exact. Idempotent set changed
  nothing; refusals wrote nothing.
- Same majestic PID throughout, 554 and 80 still listening, and a snapshot
  pulled afterwards.

**SigmaStar has not been on a camera.** Every claim is verified against vendor
source and on a host. The one thing only an SSC board can confirm is the 16-bit
RIU access: those are half-word ports in four-byte slots whose upper half is
not mapped, so a 32-bit store there writes two bytes that do not exist. I would
like an SSC30KQ or SSC338Q run before this is cut into a release.

## Follow-up, not in this PR

majestic needs a PR to call `ipchw_padmux_get`/`_set` instead of composing its
own writes, and to implement the `sdk_padmux_*` seam for sstar and ingenic —
today only `src/hisi/higpio.c` does. Until then the pin page populates
correctly on both families and writes through the old path are inert rather
than wrong, which is the fail-safe the flags above are shaped around.


widgetii:
All five addressed in 90e8f21, with a test for each. Every one was on the write
path — the half that had no camera under it when this went up. Thanks.

**1. Ethernet pads appear free to drive** — real, and the diagnosis is right,
though the fix is elsewhere. Hiding the pad is not it: infinity6c genuinely
*does* have rows saying "PAD_ETH_RN is GPIO" — six fields across three banks —
and when they are asserted the pad really is free. The bug is that `sstar_get()`
concluded GPIO from the *absence* of a known claim without ever looking at
them. Where a part has explicit GPIO fields they are now checked, and a pad
that fails them answers 0 ("exists, and this build cannot say") rather than
GPIO. That fixes the Ethernet case precisely and makes `get()` strictly more
accurate on every pad with dropped modes, without costing the pads that are
genuinely GPIO-only.

**2. Numeric pin selection fails on Ingenic** — correct, fixed. The
`IPCHW_PADMUX_F_RMW` test was guarding the wrong thing: a selector value is a
perfectly good way to *name* which function is meant, and whether writing it
takes one register or four is the backend's business. Confirmed on a T31:
`gpio mux 79 1` now muxes PC15 to UART0_CTS and back with the registers
byte-exact.

**3. Numeric muxing selects the wrong mode** — real, fixed, though the example
is not. `ETH_MODE` is not among PAD_GPIO0's alternatives on Infinity6B0; the
actual collision on that pad is **0x40, which is both `SPI1_MODE_4` and
`TTL_MODE_1`**. Same bug, same fix: all rows are scanned, and a value that
answers to more than one function name is refused with both names rather than
muxed to whichever came first.

**4. Reserved pad states can be selected** — correct, fixed. `hisi_set()` now
skips `"reserved"` exactly as `hisi_walk()` does, so the selector stays
unresolved and it returns `IPCHW_PADMUX_NO_FUNC` without writing.

**5. Mistyped pad names reconfigure another pad** — correct, fixed with `%n`,
which now has to land on the terminator. `5_2junk`, `1_2_3`, `42junk` and `5_8`
are all refused on the camera.

Re-verified on the T31 after the fixes: the malformed specs refused, the
numeric mux round trip byte-exact, selector 2 refused because PC15's FUNCTION2
is a hole, the name `reserved` refused. majestic untouched throughout and a
snapshot pulled afterwards.

SigmaStar hardware verification is deliberately postponed — the one thing only
an SSC board can settle is the 16-bit RIU access.

PR #193, opened 2026-09-15
sstar: what a pad is carrying, and how to put it back

widgetii:
Follow-up to #192, both commits found on hardware. The SigmaStar half of that
PR had never been on a camera; it has now been on seven.

## 1. An idle pad is assignable, not unknowable

The Ethernet fix in 543f5a9 asked every SigmaStar pad's "this pad is GPIO"
field before concluding GPIO. That is right for a pad whose only mode the
generator could not represent — infinity6c's multi-bank ETH and USB modes —
and wrong for everything else, which is most of the part.

On an SSC377D it hid **34 of 86 pads** behind `?`. `PAD_UART1_RX` is the shape
of it: five modes in the table, none asserted, UART1 routed nowhere
(`0x1f2079b4[6:4]` reads 0), the console on `ttyS0` somewhere else entirely — a
pad sitting idle that anybody could take. Its GPIO-mode bit is clear because
nothing has claimed it for the GPIO block yet, and reporting that as "this
build cannot say" tells a pin page to leave alone a wire that was free all
along.

The field is only worth asking where there is nothing else to go on:

| pad | what we know | answer |
|---|---|---|
| all modes in the table, none asserted | complete | GPIO — the fallback the part has |
| no modes at all | its mode was dropped | ask the GPIO field; `?` if it disagrees |

Unresolved goes from 39 pads to 5, and the 5 are exactly `PAD_USB_CID` and the
four Ethernet pads — exactly the pads whose modes the generator dropped. No
change on infinity6/6b0, which has no GPIO fields at all.

## 2. get() can say how to put a pad back

`gpio_func` is -1 on every SigmaStar row, because plain GPIO there is the
absence of every claim rather than a value. `ipchw_padmux_set(pad, "GPIO")` has
always handled that, but anything reasoning about a GPIO *selector* — a caller
that composes its own register write, as majestic's pin page does on HiSilicon
— finds nothing to reason about on that family.

The table cannot answer. `ipchw_padmux_get()` can, because it has just read the
registers: if exactly one field is claiming the pad, clearing that one field is
the whole job, and `(address, func_mask, gpio_func)` is a single write like
anywhere else. Two conditions have to hold — only one positive claim, and no
separate "this pad is GPIO" field to assert as well. **infinity6/6b0 has none
at all, so every pad there qualifies**; nearly every infinity6c and infinity6e
pad has one and keeps -1, because putting it back really does take two writes.

The lookups still answer -1, and that is not a shortcoming to fix later: they
have read nothing, so they genuinely do not know.

## Hardware verification — the SigmaStar run #192 deferred

Seven lab cameras: **SSC325DE, SSC325 (infinity6)**, **SSC335, SSC337 ×3
(infinity6b0)** and **SSC377D (infinity6c)**. majestic running throughout on
all of them, snapshot pulled from each afterwards. No infinity6e board, so
SSC338Q/SSC30KQ stays the one unverified family.

- `reginfo` byte-identical to the stock binary except the pad registers' bit 0,
  which is the live IN level and flips on the stock binary too.
- `reginfo --pads`: 85 pads on every infinity6/6b0 board, 0 unresolved; 86 on
  the SSC377D, 5 unresolved. Decoding the live mode registers independently
  with `devmem … 16` and a separate script agrees **85/85 and 86/86**.
- Corroborated per board: console on `ttyS0` ↔ the FUART pads carrying
  `UART0_MODE_2`; `I2C0_MODE_1` on `PAD_I2C0_*` (6b0) versus `I2C0_MODE_2` on
  `PAD_PWM0/PWM1` (6c — a different board layout, and the live field really
  reads 2); the MIPI sc2336 ↔ `PAD_SR_IO03..08` on `MIPI_PAD_IN_2`; the boot
  SPI-NOR ↔ the PM SPI pads on `SPI_GPIO_0`/`SPIHOLDN_MODE_0`, which are the
  **zero-valued** modes and so exercise the two-pass read on real silicon.
- `ssc325`/`ssc325de` are Macaron (`INFINITY6`) running off the I6B table,
  confirming the generator's finding that the two vendor files are identical.

**The 16-bit RIU question, which #192 said only an SSC board could settle.** A
32-bit read returns the half-word zero-extended (`0x0020` vs `0x00000020`), so
reads are safe either way. The write path was checked on `0x1F203C1C`, which
carries two adjacent PWM mode fields, on two pads that were GPIO inputs driving
nothing:

```
0x0000  ->  gpio mux 0 PWM0_MODE_4  ->  0x0004   [2:0]
        ->  gpio mux 2 PWM2_MODE_4  ->  0x0104   [8:6] set, [2:0] SURVIVED
        ->  gpio mux 2              ->  0x0004   only its own field cleared
        ->  gpio mux 0              ->  0x0000   byte-exact restore
```

— the in-place mask, the 16-bit read-modify-write and the clear sweep all
proven at once, with `get()` tracking every step.

And for the second commit, on the SSC337, a pad carrying PWM0_MODE_4 comes back
as `address=0x1f203c1c func_mask=0x0007 func=0x4 gpio_func=0 flags=RMW`, and
writing `(old & ~0x7) | 0` there **by hand with devmem, no ipctool involved**
puts the pad back to GPIO — which `padmux_get()` and `reginfo --pads` then both
agree on. On the SSC377D the equivalent pads keep -1, as they should.

## Tests

`reginfo_test` gains: a pad with modes and an unasserted GPIO field still reads
GPIO (alongside the existing case where a pad with no modes and a disagreeing
field reads `?`); and the `gpio_func` round trip — set a mode, read the row
back, compose the restore write by hand from it, confirm the pad is GPIO again,
plus the two cases that must keep -1 (a second claim asserted, and the lookups,
which have read nothing).


widgetii:
Fixed in bfd42c0. The finding is right and I had reasoned myself into
accepting exactly this gap — the recommended remedy, though, would have been
worse than the bug.

**The remedy as written re-hides more than the PR un-hides.** 57 of the
SSC377D's 86 pads have lost a claim to `OTP_TEST_1` or `TEST_IN_MODE_2` while
keeping four to thirteen named modes each. Marking every pad with a dropped
mode incomplete and falling back to its GPIO field puts all 57 behind `?` —
strictly worse than the 34 the first commit fixed, and for the same reason:
those pads' GPIO-mode bits are clear because they are idle, not because
something owns them.

**Losing the name does not have to mean losing the question.** The generator
now keeps the *field* of a dropped claim against the pad it was dropped from,
and `get()` checks those before concluding GPIO. One reading back its value
means the pad is carrying something this build can no longer name — a 0, not a
GPIO row. That is your "consult after a mode miss whenever the pad is
incomplete", done by testing the actual claim rather than by proxy, so it
costs none of the 53 pads their answer.

Only claims selected by a **non-zero** value are kept: one selected by zero
cannot be told from an idle register, and the pads that have those are the
Ethernet and USB ones the `nmodes == 0` GPIO-field check already covers. 66
fields across 53 pads on infinity6c; nothing at all on infinity6/6b0, where no
mode is dropped.

One collision the round-trip test found and nothing else would have: on the
SAR pads the vendor lists the same field *and value* under both `GPIO_MODE`
and `OTP_TEST_1`. A field that already means "this pad is GPIO" cannot also
mean "something unnameable is on it", so the generator skips it — otherwise
setting those pads to GPIO read back as unnameable.

Re-measured on the SSC377D and an SSC337 after the change: `reginfo --pads` is
byte-identical to the run before it, still 5 unresolved and still exactly
`PAD_USB_CID` and the four Ethernet pads. The difference is that the other 53
are now checked rather than assumed. Regression test added for a pad that
keeps modes and has a dropped one, with its field asserted and then cleared.

PR #194, opened 2026-09-15
ingenic: pad-mux for T21 and T23

widgetii:
Follow-up to #192/#193. Both parts answered "Platform is not supported" to
`reginfo` and to `reginfo --pads`, because the only Ingenic table was T31's.
Three lab boards — a T21N and two T23N — now answer **97, 67 and 67 pads**.

## Where the names come from, and why they are weaker

T31's came from its GPIO spec, which tables `FUNCTION0..3` for every pad and
names each *wire*. Neither of these parts has such a document anywhere in the
vendor releases, so the source is the kernel's board file,
`arch/mips/xburst/soc-<x>/include/mach/platform.h`. Two things to know before
reading the output:

- it names a **device group**, not a wire — the whole DVP bus is fifteen pads
  all called `DVP_PORTA`;
- it only covers what the board file wires up: 97 of T21's 192 pads, 67 of
  T23's 96.

Still clearly worth having. Decoded against the live registers, **80 of T21's
97** and **26 of T23's 67** sit on a function it can name, and the names are
checkable against the board itself: `I2C0_PORTA` on the two pads carrying the
sensor bus, `SFC_PORTA` on the four the SPI-NOR boots from (`mtdparts=jz_sfc`),
`MCLK_PORTA` on the sensor clock, `DVP_PORTA` across the T21's parallel sensor
bus (it runs an sc2235).

## Four things the board files get wrong

None of which may reach the table:

| | |
|---|---|
| T23 claims `PORT_D` and `PORT_F` | it has three ports. The count comes from the SoC's own enum — three for T23 and T31, six for T21, which its kernel confirms by registering six gpiochips — not from the claims |
| T23's `ssi0-pb-3-29` is `0xf << 29` on a 32-pin port | the pins past the end are dropped, and the header says so |
| some claims are a GPIO state, not a device function | a board bringing a pin up in a known state is not a mux; skipped |
| T21 has `MII_1` and `DPU_SLCD_8BIT` on one pad and one function code | the silicon cannot do both and the file does not say which is real, so that slot is given up and the header says so |

Names come from the **macro identifier** rather than the `.name` inside it,
because `.name` is not unique: T21 spells `GMAC_PORTB`,
`GMAC_INTERNAL_PHY_PORTB` and `GMAC_INTERNAL_PHY_PORTB_NO_ACTLED` all as
`"gmac_pb"`, on different function codes of the same pads. A pad offering one
device twice under one name is a pad nothing can mux deterministically — which
the existing table sweep catches, and did.

## A comment corrected, because a T23 disproved it

`PZGID2LD` is **not** what commits the four writes. Clearing bit 22 of `PBMSK`
and `PBPAT1` through their clear aliases moves both registers immediately, with
nothing written to `PZGID2LD` at all. It is still written — the vendor writes
it and it is harmless — but the comment claiming it made the four bits land
together was wrong, and with it the claim that a pad never passes through an
intermediate nibble. It does: GPIO input (`0110`) to function 0 (`0000`) is
briefly function 2. Ordering cannot fix that in general, because parking the
pin in the GPIO block first only moves the transient onto the pads that start
on a peripheral. Anyone who needs glitch-free muxing wants the shadow group
`PZGID2LD` is named for, and a TRM to work out how it is driven; this code does
not use it, and now says so.

## Verification

Hardware, on three boards, majestic running throughout and a snapshot pulled
from each afterwards:

- `reginfo` goes from a refusal to 144 lines on all three.
- `reginfo --pads` lists 97 / 67 / 67 with 4 / 10 / 10 unresolved. Decoding the
  live `PxINT`/`PxMSK`/`PxPAT1`/`PxPAT0` independently agrees **97/97, 67/67
  and 67/67**.
- The write path round-trips on a T23 — `PB22`, a GPIO input whose only
  alternative is UART0 on a board whose console is `ttyS1` — moving exactly
  bit 22 of `MSK` and `PAT1` and reverting byte-exact.
- **T31 is unchanged**: still 77 pads, 0 unresolved, same measured rows
  (`PB19=TDI`, `PB23/24=UART1_TXD/RXD`), and its generated table is
  byte-identical.

Host: native, arm32, mips32 and arm64 build clean; the knob matrix including
`IPCHW_PADMUX=none`; `reginfo_test` (which gains T21 and T23 in its sweep plus
measured cases for the six-port arithmetic and the phantom ports);
`test_pipeline.sh`; and `--verify` against the three sources. +8 KB on arm32.

T10/T20/T30/T40/T41 still have no table. T40 is the next easiest — it is the
only other T-series part with a GPIO document (`T40GPIO.xlsx`).


widgetii:
All four fixed in 24af2f8, and all four were mine from the commit before.

**1. New chips lose 28 value bits** — correct. `reginfo` on a T23 printed
`PBMSK` as `0x3` where the register holds `0xde7e003f`. I routed T21 and T23
through the Ingenic dump and forgot `dump_mask()` entirely. These are whole
controller registers, not selectors. Now `0xffffffff` for all three, confirmed
against `devmem` on both boards.

**2. The six-port chip loses half its dump** — correct, and the comment I wrote
excusing it does not survive contact with the point you make: `--pads` reads
the mux, not the controller, so nothing covered ports D–F. T21 now has its own
list, six ports plus the PZ group, 279 rows. 144 → 279 lines on the board, with
`PDMSK` and the rest present. Ports B and C of the three-port list spell a few
registers without their leading P; that is how they were entered and what
`reginfo` has always printed for T31, so they stay and the new list is
internally consistent instead of importing the typos.

**3. A conflicted pad exposes a false option** — correct, and the sharpest of
the four. The tombstone and the empty marker were the same string, so a later
claim refilled the slot: PD27 advertised `DPU_PORTD_SLCD` on function 1 while
the header four screens above recorded that exact slot as unresolvable. The
header contradicted its own data. Conflicts are now a distinct marker that
nothing later can win, emitted as `reserved`; PD27 function 1 is `reserved`.

**4. Later chips inherit old errors** — correct. Latent in the invocation I
use, since the spec-backed T31 runs first, but a latent bug is still one. Both
parsers clear the lists now.

Verified on hardware after the fixes: T21 279 dump lines with D/E/F present and
`PAMSK` = `0x440003` against `devmem`'s `0x00440003`; T23 `PBMSK` =
`0xde7e003f`; T31 `reginfo` byte-identical to the stock binary once the live
`PIN` rows are set aside, and its generated pad table unchanged.

PR #195, opened 2026-09-15
ingenic: pad-mux for T40

widgetii:
Follow-up to #194. **No T40 in the lab, so this is host-verified only** — said
up front because everything else in this series was measured on a board.

## A better source than the other two got

T40 has a **pinctrl devicetree**, which neither T21 nor T23 had. Each leaf node
of `arch/mips/boot/dts/ingenic/t40-pinctrl.dtsi` is one routing — a port, a pin
*range* and a function code — and its label says which device *and* which port:

```dts
uart0_pc: uart0-pc {
        ingenic,pinmux = <&gpc 2 5>;
        ingenic,pinmux-funcsel = <PINCTL_FUNCTION0>;
};
```

So `uart0_pc` and `uart0_pa` are two names a lookup can tell apart. That is
exactly what a board file's `.name` could not do — the reason T21 had to be
keyed on its macro identifiers, after three GMAC routings all called themselves
`"gmac_pb"`.

Still device-level rather than per-wire, like the board files and unlike T31's
GPIO spec: a routing names UART0 on port C, not which of the four pads is TXD.
**100 of T40's 128 pads**, across four ports.

## The controller needed nothing

Same `GPIO_IOBASE`, same `0x1000` stride, same `INT`/`MSK`/`PAT1`/`PAT0` at
`0x10`/`0x20`/`0x30`/`0x40`. xburst2's `enum gpio_function` numbers its
constants from `0x10` rather than `0x00`, but the low nibble is the same
encoding, so the backend took the table and a port count and nothing else.

## Two things worth naming

**The GPIO spreadsheet is not a source.** `T40GPIO.xlsx` looks promising next to
T31's GPIO spec, and it is a net list for one reference board — function, net
name, direction, and which GPIO that net lands on. Not the SoC's per-pad
functions. The devicetree is.

**Six of the fifty nodes look like conflicts and are not.** `pwm_pc` covers
PC02–PC07 at function 2 while `pwm0_pc`..`pwm5_pc` name the individual pins at
the same function: one device family spelled once for the block and once per
pin. The numbered one says more about the pad, so it wins — and only when
exactly one of the two carries a digit, so `UART0_PA` against `UART2_PA` stays
a real conflict, as does T21's `MII_1` against `DPU_SLCD_8BIT`.

## The register lists stop multiplying

#194 gave T21 its own 279-row dump list; T40 would have been a third. Ports A
through F and the PZ group are now one set of rows, and each SoC's list picks
the ports it has — six for T21, four for T40. T31 and T23 keep the three-port
list they have always dumped, typos and all, because changing it would change
what `reginfo` prints for them.

## Verification

Host: the table sweep and the set/get round trip over every pad of every
compiled-in family, now including T40; the four structural invariants; native,
arm32, mips32 and arm64 builds; the knob matrix; `test_pipeline.sh`; and
`--verify` against all four sources. +10 KB on arm32.

The T21, T23 and T31 boards confirm the register-list rework changed nothing
for them — 279, 144 and 144 dump lines, same contents, same pad counts, and
T31's `--pads` unchanged.

What would settle T40 is one board: `ipctool reginfo --pads`, then decoding the
live `PxINT`/`PxMSK`/`PxPAT1`/`PxPAT0` independently and diffing, which is what
caught the real problems on every other part in this series.


widgetii:
All three fixed in 1f32ba6.

**1. Higher selector values crash generation** — correct. `PINCTL_FUNCTION<n>`
was matched with a single `\d` and the digit used straight as an index into a
four-slot pad. T40's devicetree uses 0–2 so nothing hit it, but a generator
whose whole posture is "refuse rather than guess" should not crash rather than
refuse either. It now parses the suffix once, skips anything that is not a
number, and refuses a function number a pad does not have.

**2. Regeneration instructions no longer run** — correct, and the cause is
worth more than the symptom. `.clang-format-hook-exclude` names both generated
headers, but `apply-format` only learns about that file when the **pre-commit
hook** passes it in; a hand-run `apply-format --staged` reformats them anyway.
That is what wrapped the recipe after `--soc`, and it is the same trap that
silently dropped a `chips[]` entry during the SigmaStar work.

Two fixes, because either alone would leave the trap armed: both generators now
lay the invocation out one option pair to a line with shell continuations,
short enough that nothing wants to reflow it —

```
 *   tools/gen_ingenic_padmux.py \
 *       --soc T31 --spec T31_H.3_gpio_spec.pdf \
 *       --soc T21 --spec platform.h \
 *       --soc T23 --spec platform.h \
 *       --soc T40 --spec t40-pinctrl.dtsi
```

— and `scripts/format-changed` passes the exclusions in, with CLAUDE.md and
`docs/padmux.md` pointing at it. Both headers regenerated; both `--verify`
clean afterwards.

**3. Docs call the new chip unsupported** — correct. `docs/padmux.md` now puts
the three Ingenic source kinds side by side, because they are not equally good
and do not name things the same way:

| kind | source | names |
|---|---|---|
| `spec` (T31) | the GPIO spec's port summary tables | per **wire** — `UART1_RXD` |
| `dt` (T40) | `boot/dts/ingenic/<soc>-pinctrl.dtsi` | per **device and port** — `UART0_PC` |
| `platform` (T21, T23) | `soc-<x>/include/mach/platform.h` | per **device** — `UART0_PORTB` |

It also says which parts have which, warns that `T40GPIO.xlsx` is a board net
list rather than the spec it resembles, and separates the parts verified on
hardware from the two — SigmaStar infinity6e and Ingenic T40 — that have only
ever been verified on a host.

PR #196, opened 2026-09-16
sstar: infinity6e on hardware, and the pads its table lies about

widgetii:
First SSC30KQ to hand, so the infinity6e table finally met a board — it was the last SigmaStar family that had only ever been checked on a host.

## What the board confirmed

Every register the table names decodes exactly as `reginfo --pads` reports, cross-checked against an independent `devmem` decode:

| register | raw | independent decode | ipctool |
|---|---|---|---|
| `0x1F2079C0` | `0x0004` | `SPI_MODE`, `SPIHOLDN_MODE` | pads 116–118, 120 ✓ |
| `0x1F2079BC` | `0x0113` | `I2C1_MODE_1` | pads 85/86 — the IMX335 bus ✓ |
| `0x1F20798C` | `0x0021` | `LED0_MODE_1`, `LED1_MODE_2` | pads 44/45 ✓ |
| `0x1F2079A4` | `0x0111` | `SR00_MIPI_MODE_1`, `SR00_CTRL_MODE` | pads 69–78, 82–84 ✓ |
| `0x1F007F94` | `0x0000` | `PM_PM_UART_IS_MODE` | pads 2/3, the console ✓ |
| `0x1F207994` | `0x1101` | PWM0=1, **PWM1=0**, PWM2=1, PWM3=1 | pads 107 ✓, 108 *not* PWM ✓, 109 ✓, 110 ✓ |
| `0x1F207984` | `0x0000` | `EMMC0_8B_MODE_1` **not** active | absent ✓ (NOR board) |

The four PWM fields of one register are the useful case: a decoder that ignored field offsets would report all four, and the negative is what pins it. The kernel's exported `gpio13`/`gpio14` both read `[GPIO]`, and pad 58's `[?]` is correct — its GPIO-enable bit is genuinely clear (`0x75 & 0x8 = 0`) where pad 57's is set.

## What it did not confirm

Six pads. `PAD_ETH_RN` through `PAD_USB2_DP` were reported as carrying `SPIHOLDN_MODE` and offering `EMMC0_8B_MODE_1` — i.e. eMMC data lines on the Ethernet magnetics.

`m_stPadMuxTbl` is not the whole truth. Those pads are muxed from banks the table never names — an Ethernet pair by `REG_ETH_GPIO_EN` in ALBANY2, a USB pair by the UTMI0 power-down bits — and the vendor routes exactly them through `HalPadSetMode_MISC()`. Their table rows are copy-paste from the pad above, naming **`PAD_SPI_HLD`'s own fields**. Meanwhile `REG_ETH_GPIO_EN` read `0x0000`, the link was up, and the pads were carrying Ethernet.

The generator caught the equivalent on infinity6b0 and infinity6c only because the vendor's rows *there* disagreed with each other and the multi-tuple rule dropped them. Structure cannot separate a copy-paste row from a real pad group — a mode claiming several pads through one field is exactly what `FUART_MODE_2` legitimately is. So read the dispatch instead: parse `HalPadSetMode_MISC()` for what it writes per pad, and drop a pad's rows when they share **no** register with it.

Overlap, not containment: the vendor is also inconsistent about which of a group's modes it lists, giving `PWM9_MODE_2` to `PAD_FUART_CTS` but not `PAD_FUART_RX` though MISC programs both. Requiring the table to be complete would throw away good rows. The FUART and SAR pads are muxed through their own fields either way and keep everything.

**infinity6b0 and infinity6c regenerate byte-identical**, so the seven SigmaStar cameras and the SSC377D verified earlier are untouched. The i6e diff is 12 rows and 6 pads.

## A footgun found while proving it

`gpio mux <pad>` with no function did not ask what the pad carries — it muxed the pad to GPIO. There was no query form at all, and the mistake is silent: run on `PAD_ETH_LED0` it dropped `LED0_MODE_1` and left the link LED dark until the mode was put back. (Pre-existing — `new_func = i` was the no-arg fallback long before the padmux series — but it now writes on three vendors instead of one.)

It now reports, which is also the only way to see a pad that the drop above hides from `reginfo --pads`:

```
# ipctool gpio mux 121
pad 121  -                [?]
```

Setting a pad to GPIO keeps its unambiguous spelling, `gpio mux <pad> GPIO`.

## Write path, measured

On pad 88 of the unused second sensor port, with `PWM1` unrouted so nothing else moved:

- `set(88, PWM1_MODE_4)` → `0x1F207994` `0x1101 → 0x1141`; the PWM1 field `[6:4]` went 0→4 and **nothing else in that register or the six others the pad can reach** moved. PWM0/2/3 preserved.
- `get(88)` read back `PWM1_MODE_4`.
- `set(88, GPIO)` restored the field *and* asserted the pad's GPIO bit — both halves of SigmaStar's "back to GPIO".
- Registers byte-exact to baseline afterwards; eth0 up, majestic streaming IMX335.

## Tests

`reginfo_test` gains the measured cases: the six MISC pads answer nothing from `get`/`by_pad` and refuse `set`, `SPIHOLDN_MODE` is left to the pad that owns it, and the `0x1101` four-PWM decode including the idle one.

Verified: native + `reginfo_test` + `cYAML_test`, arm32/mips32/arm64 cross builds, `test_pipeline.sh`, `IPCHW_VENDORS=sstar IPCHW_PADMUX=sstar` and `=none` trimmed configures, and `--verify` round-trips the header against the SDK.

## Unrelated, noted not fixed

`cmake --build -jN` races on the generated `version.c`: both `ipctool` and `ipcinfo` depend on a file regenerated every build, so a parallel build intermittently compiles it mid-rewrite. `-j1` is clean and CI builds without `-j`, so it is invisible there. No CMake file is touched by this PR.


widgetii:
Fixed in `306a8e0`. The finding is real — thanks.

`RIUA_RE`'s `(.+?)\)` ended the second argument at the *inner* parenthesis of the one nested offset any vendor MISC path contains — `REG_FUART_RX_GPIO_MODE+(u32PadID-PAD_FUART_RX)`, which is the form the comment beside the regex documents:

```
>>> RIUA_RE.findall("_RIUA_16BIT(PADTOP_BANK,REG_FUART_RX_GPIO_MODE+(u32PadID-PAD_FUART_RX)), BIT3, MASK)")
[('PADTOP_BANK', 'REG_FUART_RX_GPIO_MODE+(u32PadID-PAD_FUART_RX')]
```

One correction to the mechanism, since it changes which case is dangerous. A shrunken `regs` set makes `named & regs` *less* likely to be non-empty, so dropping a register biases toward **over**-dropping — losing good rows, not keeping bad ones. The way bad rows survive is the other one: `out.setdefault()` only runs for a register that resolves, so a pad whose registers **all** fail never enters the dict, `misc_conflicts()` never examines it, and its rows stay. Same conclusion, and it is the worse failure because it is the one this check exists to prevent.

Both are now impossible: arguments split on top-level commas with a balanced scan, and an expression that will not resolve **refuses** instead of being skipped. Across all three families there is exactly one nested expression and every other offset is a bare identifier, so refusing costs nothing today and is loud the day it does not.

**The tables are unchanged** — the FUART pads already overlapped through `REG_PWM0_MODE`, so the missing register never changed a verdict. All three families re-derive byte-identical, which is the check that says this was a latent bug rather than a live one.

On the recommended fix's second half, I went a step further than the parser. `--verify` needs a vendor SDK, so **CI has never run these generators at all** — there was no guard to regress. `gen_sstar_padmux.py --selftest` now parses a built-in fixture and `tools/test_pipeline.sh` runs it. Every shape in the fixture is one that has already gone wrong: the nested offset, a fall-through `case` group whose pads must each receive the group's registers with the per-pad offset resolved per pad, an unresolvable register that must refuse rather than skip, and an overlap that keeps a pad against a non-overlap that drops one. Verified both ways — it passes on this parser and fails on the one it replaces.

On the two alternatives in the summary: agreed on both counts. Hard-coding the six pads would have encoded knowledge that already exists in `HalPadSetVal()`'s dispatch and would miss the next SDK's version of this. Full ALBANY/UTMI backends are real work for pads nothing currently asks about — worth doing when a caller needs to configure an Ethernet or USB selector, and not before.

PR #198, opened 2026-09-16
hisi: put the unassigned selector values back in the V2 tables

### LINKED ISSUE #135: Incorrect pinmux values for SDIO1 on Hi3518E V200
Hello,
I'm working on a small camera with hi3518ev200 chip and a wifi module connected to SDIO1. Original stock firmware is a LiteOS clone. I've managed to flash openipc on it, and after investigating both stock firmware and ipctool readouts it seems there are discrepancies in your SDIO1 mappings in EV20Xregs array.
According to "ipctool reginfo" almost all SDIO1_ pins correspond to pinmux register value "0x3", except SDIO1_CARD_POWER_EN which is 0x4. This seems wrong, my stock firmware sets all those values to 0x4. I've replicated that and after some more shenanigans with onboard MCU (which is connected to UART1 and controls PWR and RST pins for wifi) I've finally managed to make wifi card at least present itself on MMC bus (as SDIO_ID 0296:5347 which seems to be in kernel already as Wilc1000 but is NOT actually that card or very much compatible at all :-/).
Anyway, this is pinmux configuration which makes SDIO1_* pins work correctly (at least those used on my board, I'm not sure about SDIO1_CARD_DETECT, SDIO1_CWPR and SDIO1_CARD_POWER_EN but have no reason to doubt the stock firmware setting them right):
```
muxctrl_reg4 0x200f0010 0x4 GPIO2_0 RMII_CLK VO_CLK SDIO1_CCLK_OUT
muxctrl_reg5 0x200f0014 0x4 GPIO2_1 RMII_TX_EN VO_VS SDIO1_CARD_DETECT
muxctrl_reg6 0x200f0018 0x4 GPIO2_2 RMII_TXD0 VO_DATA5 SDIO1_CWPR
muxctrl_reg7 0x200f001c 0x4 GPIO2_3 RMII_TXD1 VO_DE SDIO1_CDATA1
muxctrl_reg8 0x200f0020 0x4 GPIO2_4 RMII_RX_DV VO_DATA7 SDIO1_CDATA0
muxctrl_reg9 0x200f0024 0x4 GPIO2_5 RMII_RXD0 VO_DATA2 SDIO1_CDATA3
muxctrl_reg10 0x200f0028 0x4 GPIO2_6 RMII_RXD1 VO_DATA3 SDIO1_CCMD
muxctrl_reg11 0x200f002c 0x4 GPIO2_7 EPHY_RST BOOT_SEL VO_HS [SDIO1_CARD_POWER_EN]
muxctrl_reg13 0x200f0034 0x4 GPIO3_0 EPHY_CLK VO_DATA1 SDIO1_CDATA2
```
Some of these pins I've tested manually in GPIO mode and that seemed to work fine with pinmux reg 0x0, and the default power-on values are 0x1 which would point to RMII/EPHY also being correct, but I have no idea what exactly should be mapped to 0x2/0x3 values.
Is it possible perhaps to insert "UNKNOWN" as 0x3 in there so the SDIO1 pins are in correct position? It's cost me quite some time and frustration wondering why nothing works with original 0x3 values...
If it helps, this is basic "ipctool" output for my board:
```
root@openipc-hi3518ev200:~# ipctool
mdio busy
mdio busy
---
chip:
  vendor: HiSilicon
  model: 3518EV200
board:
  vendor: OpenIPC
  version: 2.5.04.30
ethernet:
  u-mdio-phyaddr: 0
  phy-id: 0x00000000
  d-mdio-phyaddr: 0
rom:
- type: nor
  block: 64K
  partitions:
    - name: boot
      size: 0x40000
      sha1: 3be1edb5
    - name: env
      size: 0x10000
      sha1: 623a40f5
      contains:
        - name: uboot-env
          offset: 0x0
    - name: kernel
      size: 0x200000
      sha1: c442306d
    - name: rootfs
      size: 0x530000
      sha1: c6fde31d
    - name: rootfs_data
      size: 0x80000
      sha1: 402772f1
  size: 8M
  addr-mode: 3-byte
ram:
  total: 32M
firmware:
  u-boot: "2010.06 (Nov 14 2022 - 19:23:46)"
  kernel: "4.9.37 (Fri May 9 17:50:50 CEST 2025)"
  toolchain: buildroot-gcc-13.3.0
  main-app: majestic
```
Thanks a lot!
BR,
relghuar

widgetii:
Closes #135.

`funcs[]` is indexed by the selector value, so an unassigned value has to be present as `"reserved"` or every function after it is off by one. The Hi3518EV20X and Hi3516CV200 tables had the holes closed up — **19 registers of each**, including the nine that carry SDIO1.

## The reporter was right, and the hole is one earlier than they guessed

@relghuar's stock firmware drives a WiFi module on SDIO1 and sets every SDIO1 pin to `0x4`; ipctool said `0x3`. The data sheet settles it — `muxctrl_reg4` reads:

```
000: GPIO2_0
001: RMII_CLK
011: VO_CLK
100: SDIO1_CCLK_OUT
```

`010` is unassigned. So `VO_CLK` is 3 and `SDIO1_CCLK_OUT` is 4, and all nine SDIO1 pins now match what your firmware writes. You asked whether `UNKNOWN` could go in at `0x3` — it turns out the gap is at `0x2`, so `VO_*` was displaced too, which is why `0x2`/`0x3` both looked unexplainable from the outside.

## Where the error came from

The document states this twice, and the two do not agree:

| chapter | shape | for `muxctrl_reg4` |
|---|---|---|
| 2.3 Pin Multiplexing Control Registers | the bit field, value by value | `000: GPIO2_0`, `001: RMII_CLK`, `011: VO_CLK`, `100: SDIO1_CCLK_OUT` |
| 2.4 Software Multiplexed Pins | columns headed "Multiplexed Signal 0", "Signal 1", "Multiplexed Signals **2-4**" | `GPIO2_0`, `RMII_CLK`, `VO_CLK`, `SDIO1_CCLK_OUT` |

Chapter 2.4 lists alternatives by **position**, holes closed up, and its last column is a *range* that never says which value is which. Transcribed as selector values it produces exactly the rows that were here. Where 2.4 does number them it agrees with 2.3 — `muxctrl_reg3` is written out as `2: SFC_DEVICE_MODE`, **`3: -`**, `4: VI_HS_BT1120`, an explicit dash for the hole.

## Fixed in full, not just the nine pins

Both V2 tables were diffed against chapter 2.3 in its entirety. **19 of 66 rows disagreed in each**, and every single one was a missing hole — no renames, no wrong addresses, no register in the data sheet absent from the table. That the mismatches are all one shape is itself the evidence the parse is sound.

The two SoCs' chapters are identical for all 66 registers, which is why the two tables are copies and both needed the same fix. After it, **0 of 66** disagree.

Beyond SDIO1 this also corrects `VO_*` on 10 registers, `TEMPER_DQ` on 2, `I2S_*`/`UART1_*`/`UART2_*` on 6, and `VI_HS_BT1120`.

## Reproducible

`tools/check_hisi_padmux.py` is that diff, checked in so the claim can be re-run:

```
$ tools/check_hisi_padmux.py --datasheet <ds>.pdf --prefix EV20X_ \
      --base 0x200f0000 --soc "Hi3518E V20X" --reginfo <parent>/src/reginfo.c
19 of 66 rows disagree

$ ...--reginfo src/reginfo.c        # this commit
0 of 66 rows disagree
```

It will `--fix` a missing hole but **refuses** when a function *name* differs, on the grounds that the parse is then likelier wrong than the table. It also refuses to guess when a document covers several SoCs — this one holds a 192-pin Hi3518EV20X and a 273-pin Hi3516CV200 back to back — and makes you pass `--soc`.

Data sheets are not in this repo, so CI cannot run it. **The other HiSilicon families have not been checked this way** (AV100, CV100, CV300, AV200, CV500, EV300, DV200/300/500, CV610, RCV100 …); whoever has their documents should.

## No new writable state

`reginfo.c` already treats `"reserved"` as a hole — the walks skip it, `get()` reports nothing for it, and `set()` refuses it. The new entries shift the selectors without offering anything new to write.

## One existing assertion changed

`reginfo_test` had `PWM0` on `muxctrl_reg31` at selector 3. That register is `000/001/011/100` as well, so it is 4 — the assertion was encoding the table, not a measurement. (`muxctrl_reg58`, the other `PWM0`, has no hole and its `func == 2` is unchanged.)

The V2 case now also pins all nine SDIO1 pins at 4 and checks that `"reserved"` is never offered as a function of a pad. Reverting any single row fails it — checked, with a non-zero exit.

## Verified

Native + `reginfo_test` + `cYAML_test`, arm32/mips32/arm64 cross builds, `test_pipeline.sh`. arm32 `.text` grows 148 bytes (the 38 extra pointers; the string is shared). No hardware — I have no EV200 board, so the confirmation is the data sheet plus @relghuar's own register readout, which agree on all nine SDIO1 pins.


widgetii:
All four fixed in `78c7bfb`. Every one was real, and they share a shape worth naming: **a check that stays quiet instead of failing.**

**1, 2, 3 — `check_hisi_padmux.py`.**

A register present in the data sheet with no row at all was printed and then forgotten — it never reached the count or the exit status, so a table missing rows could report `0 of N rows disagree` and exit 0. Worth stating plainly: a pad with no row is absent from *every* lookup, which is a quieter wrong answer than a shifted selector, not a smaller one.

`--fix` returned 0 unconditionally, so a run that rewrote nothing — because every mismatch was of a kind it refuses to touch — still claimed rows were rewritten and exited clean.

And `(addr - base) // 4` floored, so a row at `base + 0x11` aliased register 4 and could pass on another register's functions. The table's address is what gets read and written, so that is not a naming quibble.

The three share one cause, so they share one fix: mismatches are now split into **missing holes**, which the tool knows how to repair, and **everything else** — no such register, bad address, renamed function, register with no row. If anything lands in the second bucket, `--fix` writes **nothing**. That is stronger than the recommended fix, deliberately: each of those is at least as likely to mean the wrong document, `--soc` or `--base` as a wrong table, and repairing a table against the wrong document would wreck it rather than improve it.

That also removed a smaller bug you didn't flag: a renamed function used to abort the run via `sys.exit()` *after* earlier rows had already been rewritten in memory. Now every problem is reported before anything is decided.

**4 — `one()` in `reginfo_test.c`.** Correct, and it applies to the four pre-existing call sites as well as my new loop. Every call site sits inside the `#ifdef` for that family's table, so the function is always meant to be there — not finding it is the failure, not a reason to stop looking. `one()` now records a failure, which is what makes the SDIO1 loop mean anything: renaming `SDIO1_CCLK_OUT` in the table used to pass, and now fails.

**Checked by breaking each one**, since a guard never seen to fail isn't known to be a guard: a deleted row (`rc=1`), an unaligned address (`rc=1`), a hole plus something unfixable (file left byte-identical, `rc=1`), a hole alone (rewritten, re-check clean, `rc=0`), a renamed function (reported, file untouched, `rc=1`), and a function removed from the V2 table (`FAIL expected 1 pad for SDIO1_CCLK_OUT, got 0`, exit 1).

One of those first ran with a `str.replace` that matched nothing and quietly proved nothing — the two V2 tables are now identical, so an anchor taken from a row *body* appears twice. I nearly recorded a pass off it. Assert on the replacement count.

The tables themselves are unchanged by this commit: both still report `0 of 66 rows disagree`.

widgetii:
Correcting the "beyond SDIO1" figures in the description — I mixed register counts and function counts there, and left `PWM0` out. Computed from the diff, these are the **distinct functions whose selector value changed** (identical in both tables):

| group | n | functions |
|---|---|---|
| `VO_*` | 10 | `VO_CLK` `VO_DATA0` `VO_DATA1` `VO_DATA2` `VO_DATA3` `VO_DATA4` `VO_DATA5` `VO_DATA7` `VO_DE` `VO_VS` |
| `I2S_*` / `UART*` | 9 | `I2S_BCLK_TX` `I2S_MCLK` `I2S_SD_RX` `I2S_SD_TX` `I2S_WS_TX` `UART1_CTSN` `UART1_TXD` `UART2_RXD` `UART2_TXD` |
| `SDIO1_*` | 8 | `SDIO1_CCLK_OUT` `SDIO1_CARD_DETECT` `SDIO1_CWPR` `SDIO1_CDATA0..3` `SDIO1_CCMD` |
| other | 3 | `PWM0` `TEMPER_DQ` `VI_HS_BT1120` |

30 in total. Two things worth drawing out:

**Eight SDIO1 functions moved, not nine.** `SDIO1_CARD_POWER_EN` sits on `muxctrl_reg11`, whose value 2 is `BOOT_SEL` — a real function, not a hole — so that row was already correct. That is exactly what @relghuar saw from the other side: *"almost all SDIO1_ pins correspond to pinmux register value 0x3, except SDIO1_CARD_POWER_EN which is 0x4"*. The one pin that looked anomalous was the only one that was right, which is a good tell for this bug class generally — a lone row disagreeing with its neighbours usually means the neighbours lost a hole, not that the row is odd.

**`PWM0` moving is what forced the test change**, so it belongs in this list rather than only in the note further down.

PR #199, opened 2026-09-16
hisi: check the rest of the families, and fix two holes in AV100

widgetii:
Follow-up to #198, which fixed the two V2 tables and left every other HiSilicon family unchecked. Checking them turned out to be mostly a question of whether the document exists.

## Result

| table | SoC | rows | result |
|---|---|---|---|
| `CV100` | Hi3518E V100 | 87 | **agrees** — on both document revisions |
| `AV100` | Hi3516A/D V100 | 123 | **2 holes restored** — both revisions agree |
| `EV20X` | Hi3518E V20X | 66 | 19 holes restored (#198) |
| `CV200` | Hi3516C V200 | 66 | 19 holes restored (#198) |

```
muxctrl_reg54   00: GPIO0_1               10: TEMPER_DQ
muxctrl_reg122  00: GPIO13_7  01: VI_DAT0            11: PWM4
```

Both skip a value, so `TEMPER_DQ` is selector **2** and `PWM4` is **3**; the table had them at 1 and 2. Nobody reported these — they came out of diffing all 123 rows.

`TEMPER_DQ` is the SDIO1 case from #198 read from the other side. It sits at selector 2 on `muxctrl_reg53`, `55`, `56` and `57` as well, each of those with a real function at 1 — so `reg54` offering it at 1 was the single row out of step with its neighbours. That asymmetry is the tell for this whole bug class, and it is what `SDIO1_CARD_POWER_EN` was in #135.

## Cross-checked against a second revision

Both findings hold on **both** revisions of the Hi3516A/D data sheet, SPC050 and SPC080. Their only disagreement is a name: SPC050 calls `muxctrl_reg93` value 2 `RMII_CLK`, SPC080 writes it out as `RMII_CLK_OUT/MII_TX_CLK`. The table carries the newer spelling and the selector is the same either way — and the checker's refusal to touch a *renamed* function is exactly what kept it out of the diff. First time that guard has earned itself on real data rather than a synthetic break.

## Why the rest cannot be checked

Not for want of looking. **From Hi3516CV300 onwards the data sheet stopped carrying chapter 2.3 entirely.** The newer documents mention `iocfg_reg55` in passing and define none of them, so the encoding lives in a pin-multiplexing document that is not in these SDK releases. I checked every Data Sheet in the SDK tree, including older releases of the same SoCs.

That leaves `AV200`, `CV300`, `CV500`, `DV300`, `EV200`, `EV300`, `_8EV300`, `DV200`, `DV500`, `CV610`, `RCV100`, `DV100` — about **900 rows** — unchecked. SDK source is not a substitute: it shows the values a driver *writes*, and a hole is exactly the value nobody writes.

If anyone has a HiSilicon "pin multiplexing" document for those parts, the tool takes it directly.

## Three parser fixes

Each was found by a table that disagreed in a way that turned out to be the tool's fault, not the table's:

- **A register's name is repeated inside its own bit-layout diagram**, and in the Hi3518EV100 sheet that copy is pushed out to column ~140 and comes back from `pdftotext` with its trailing digits clipped — `muxctrl_reg22` reads as `muxctrl_reg2`. Keying on it handed one register's values to another five pages earlier, and made CV100 look like it had four missing registers and one wholly wrong row. Section headings start at column 0; only those count now.
- **`10: RMII_CLK_OUT/MII_TX_CLK`** is one value with two names, and the tables carry the slashed spelling verbatim. The value pattern stopped at the slash, so all 17 such rows read as renames.
- **`11: reserved` written out** is the same statement as leaving `11` off the list. Recording it as a function made rows read one entry too long.

**EV20X and CV200 still report `0 of 66` after all three** — which is what says these are parser fixes and not new behaviour.

## Tests

New `test_v2a_holes` pins both corrected rows plus the slashed name. It looks them up *by address*, because `TEMPER_DQ` is on five pads here and `PWM4` on two — `one()` is the wrong helper for this table. Reverting either row fails it; checked both, with non-zero exits.

Verified: native + `reginfo_test` + `cYAML_test`, arm32/mips32/arm64, `test_pipeline.sh`, and all four checkable tables reporting `0 disagree`.


widgetii:
One addition since the review, in `88aeb8c` — docs only, no code.

Every table checked so far has now been diffed against **two independent revisions** of its own data sheet, not one. That includes the 19+19 rows fixed in #198, which I had only ever run against a single document:

| table | rows | result | revisions |
|---|---|---|---|
| `CV100` | 87 | agrees | SPC081, SPC0B0 |
| `AV100` | 123 | 2 holes restored | SPC050, SPC080 |
| `EV20X` | 66 | 19 holes restored | SPC040, **SPC050** |
| `CV200` | 66 | 19 holes restored | SPC040, **SPC050** |

All four agree on every selector value across both revisions. The single difference anywhere in ~340 rows is a name — `muxctrl_reg93` on AV100 is `RMII_CLK` in SPC050 and `RMII_CLK_OUT/MII_TX_CLK` in SPC080 — and the checker's refusal to touch a rename is why it never reached the diff.

Worth stating because the #198 fix was a 38-row change derived from one document. It now has a second source behind it.

Issue #138, opened 2025-10-19
Cameras USB


AlvaRoBot7:
Can someone help me? I have some cameras but I don't know how to identify them. I don't know if I can use IPCTool and how to use it.

Big111156:
Have the same Issue!...

Neither on Linux Mint (64bit) nor on Windows 11 (64bit) got it working...
Pls more Information... Thanks

openipc-ai:
`ipctool` is not a desktop program — it is a small static binary that you copy **onto the camera** and run **there**. It reads the camera's own SoC registers (`/dev/mem`), its I2C/SPI buses, `/proc` and the MTD flash partitions, so on a PC there is simply nothing for it to find.

That is why nothing happened on Linux Mint or Windows 11:

* there is no Windows build at all — the release assets are Linux static binaries for `arm32` (`ipctool`), `mips32` (`ipctool-mips32`) and `arm64` (`ipctool-arm64`);
* on an x86 Linux desktop it does compile and start, but SoC detection fails and it exits **silently** with status 1. (I just reproduced that on a PC: no output, `echo $?` → 1. The silence is a rough edge on our side — it should tell you *why* it gave up.)

### How it is actually used

1. **Get a shell on the camera** — telnet, SSH, or the UART console. The README also lists NFS / TFTP / `uget` / `telnet_upload.py` tricks for cameras with no working network file transfer.
2. **Check the camera's CPU**, not your PC's: `cat /proc/cpuinfo` on the camera.
   * ARMv5/ARMv7 (HiSilicon, Goke, SigmaStar, Xiongmai, Fullhan, Novatek…) → `ipctool`
   * MIPS (Ingenic T10…T41) → `ipctool-mips32`
   * aarch64 (newer HiSilicon, e.g. Hi3519DV500) → `ipctool-arm64`
3. **Copy it to `/tmp`** — that is tmpfs and usually the only writable place with room — then `chmod +x /tmp/ipctool` and run it **as root**.
   Most cameras ship no `sftp-server`, so plain `scp` fails; use `scp -O ipctool root@CAMERA_IP:/tmp/`, or download it on the camera itself from the release URL in the README.
4. `./ipctool` with no arguments prints the full YAML report. `ipctool -c`, `-s`, `-t` are the one-line forms (chip / sensor / temperature).

### If you meant USB webcams

[...]

Issue #134, opened 2025-01-05
ipctool not working on Victure PC420


rira0815:
Hello,

I have a Victure PC420 camera and I already got telnet and ftp working on it.
On the camera I tried to collect Information with the ipctool, but it seems not to work.

When I type in `./ipctool` the output is nothing

When I type in `./ipctool -h` the output is the "Usage" text.

output from dmesg on the camera:

```
Booting Linux on physical CPU 0
Linux version 3.4.35 (zhout@apemans_server) (gcc version 4.8.5 (anyka (gcc-4.8.5 + binutils-2.24 + ulcibc-0.9.33.2)(20170223)) ) #9 Thu Feb 25 11:45:35 UTC 2021
CPU: ARM926EJ-S [41069265] revision 5 (ARMv5TEJ), cr=00053177
CPU: VIVT data cache, VIVT instruction cache
Machine: CLOUD39EV3_AK3918EV300_MNBD
Memory policy: ECC disabled, Data cache writeback
ANYKA CPU AK3918 (ID 0x20160100)
On node 0 totalpages: 9472
free_area_init_node: node 0, pgdat c03e9084, node_mem_map c0411000
  Normal zone: 74 pages used for memmap
  Normal zone: 0 pages reserved
  Normal zone: 9398 pages, LIFO batch:1
pcpu-alloc: s0 r0 d32768 u32768 alloc=1*32768
pcpu-alloc: [0] 0 
Built 1 zonelists in Zone order, mobility grouping on.  Total pages: 9398
Kernel command line: console=ttySAK0,115200n8 root=/dev/mtdblock4 rootfstype=squashfs init=/sbin/init mem=64M memsize=64M
PID hash table entries: 256 (order: -2, 1024 bytes)
Dentry cache hash table entries: 8192 (order: 3, 32768 bytes)
Inode-cache hash table entries: 4096 (order: 2, 16384 bytes)
Memory: 37MB = 37MB total
Memory: 33276k/33276k available, 4612k reserved, 0K highmem
Virtual kernel memory layout:
    vector  : 0xffff0000 - 0xffff1000   (   4 kB)
    fixmap  : 0xfff00000 - 0xfffe0000   ( 896 kB)
    vmalloc : 0xc2800000 - 0xff000000   ( 968 MB)
    lowmem  : 0xc0000000 - 0xc2500000   (  37 MB)
    modules : 0xbf000000 - 0xc0000000   (  16 MB)
[...]

PR #200, opened 2026-09-16
cli: say why detection failed instead of printing nothing

openipc-ai:
An unrecognised SoC made `ipctool` exit without a word.

`getchipname()` returns NULL when `hw_detect_system()` matches neither a known UART0 base in `/proc/iomem` nor a `manufacturers[]` pattern against the `/proc/cpuinfo` `Hardware`/`vendor_id`/`machine` line. `build_yaml()` then returned NULL and `main()` printed an empty report and exited 1. Because `-h` still works — it touches no hardware — this reads as a broken binary rather than as unsupported hardware. That is exactly how it gets reported:

* #134 — a Victure PC420, i.e. an Anyka AK3918EV300; Anyka has no HAL here at all.
* #138 — the ARM binary run on an x86 desktop.

`explain_unknown_chip()` names the failure and echoes back the two inputs detection actually looked at, so a bug report carries them without anyone having to ask:

```
ipctool: SoC not recognised, no hardware report possible.
  no UART0 entry in /proc/iomem (reading it needs root)
  /proc/cpuinfo reports: GenuineIntel
This is an x86 build. ipctool probes camera hardware
directly, so it has to run on the camera itself, not on a
desktop.
```

Details:

* stderr only, so stdout stays empty and `ipctool -c` in a script behaves as before;
* the closing advice differs under `#if defined(__i386__) || defined(__x86_64__)`, where the answer is not "file a bug" but "run it on the camera";
* it sits behind `#ifndef STANDALONE_LIBRARY`, so `libipchw` stays silent;
* the i2c/spi paths printed a bare `Unknown chip` for the same condition and now share the message, and `i2cdetect` printed its table header before finding out, so that moves after the check.

### Verification

* native Release build clean, no new warnings in the touched files;
* `cYAML_test` and `reginfo_test` pass;
* `-DONLY_LIBRARY=ON` builds clean, confirming the `STANDALONE_LIBRARY` guard;
* both arms of the x86/non-x86 `#if` compiled and inspected by temporarily forcing the branch;
* no cross toolchain on this machine, so arm32/mips32/arm64 rest on CI; the change is arch-neutral apart from that one `#if`.
* `clang-format` was not available locally, so the lines were hand-kept inside the LLVM 80-column limit rather than run through `scripts/format-changed`.


openipc-ai:
Both review findings were real; fixed in 182f331.

**1. "Failed detection reports unused data."** Correct, and the fix turned out to be worth more than tidiness. The two detection paths are alternatives — a known UART0 base picks a vendor detector and never reads `/proc/cpuinfo`, while only the fallback walks `manufacturers[]` against it — so `detect_and_set()` and `generic_detect_cpu()` now record which one ran, and the diagnostic reports that one. That also separates two failures worth telling apart: a *known* UART0 base whose chip never identifies itself is a supported SoC whose identity register did not read back, which is exactly what `CONFIG_STRICT_DEVMEM` looks like from here, so it now says so:

```
  UART0 base 0x12100000 in /proc/iomem is a known one, but the
  chip did not identify itself -- reading its registers
  through /dev/mem may be blocked (CONFIG_STRICT_DEVMEM)
```

I kept the state as two file statics in `chipid.c` rather than threading a result struct through `hw_detect_system()` and its callers, per the PR's own "minimize API churn" reasoning.

**2. "Unreadable UART data is called absent."** Also correct. `explain_missing_uart0()` now separates an unreadable `/proc/iomem` (reported with `strerror(errno)`) from a readable one with no match, and in the common case — not root, so `/proc/iomem` zeroes every address — says that instead of leaving the reader to guess:

```
  no UART0 entry in /proc/iomem, which zeroes every
  address unless you are root -- try again as root
```

I did not widen `line_from_file()` to carry a reason as suggested: it has callers all over the tree and only this path wants the distinction, so the diagnostic makes it locally.

Verified again after the change: native build clean, `cYAML_test` and `reginfo_test` pass, `-DONLY_LIBRARY=ON` builds with no warnings (the new statics are written in the library build and read only behind `#ifndef STANDALONE_LIBRARY`), and all four message branches — known base, unknown base, missing `/proc/iomem`, plus both arms of the x86 `#if` — were compiled and inspected by forcing each in turn.

PR #201, opened 2026-09-16
tools: fixtures for the HiSilicon checker, and a check that CI runs them

widgetii:
Follow-up to #199, taking up the suggestion from its review.

`check_hisi_padmux.py` had no test of its own. Its `--verify` needs a data sheet, which cannot live in this repo, so **CI exercised none of it** — and three of its parsing rules were fixed in #199 only because a table disagreed loudly enough to be worth investigating. A quieter version of any of them would have gone straight into a generated diff.

## The fixture

`--selftest` runs the parsing and the row verdicts against a built-in fixture, written in the shape `pdftotext -layout` actually emits — because the layout is the hard part, not the content. Every case in it has silently produced a wrong answer at least once:

| case | why it is there |
|---|---|
| value list split by a page break, hole in the middle | the shape the whole bug class lives in |
| single-bit field | demanding three binary digits drops every 1- and 2-bit register in the document |
| `10: FFF_OUT/GGG_CLK` | one value, two names; truncating at the slash made 17 real rows read as renames |
| `11: reserved` spelled out | a hole, not a fifth function — recording it made rows one entry too long |
| bit diagram whose clipped label reads `muxctrl_reg2` but means `muxctrl_reg22` | handed one register's values to another twenty registers earlier, and made CV100 look like it had four missing registers and a wrong row |
| two SoCs' chapters in one document | Hi3518EV20X and Hi3516CV200 ship back to back |
| a value too wide for its own field | a parse error, and one that has to be said rather than published |

**Checked by reverting each of the four parsing rules in turn.** Every one fails the selftest, with a message naming the register and what it got — e.g. `reg2 (slash, trailing reserved): ['GPIO0_2', 'EEE', 'FFF_OUT']`.

## A small extraction

The per-row decision moved into `classify()`, which is what decides whether `--fix` may touch a row, and now returns a verdict instead of printing one. Behaviour is unchanged and I checked that against every document I have: `EV20X`, `CV200`, `CV100` and `AV100` all still report `0 of N`, and `AV100` still reports exactly one row against the older SPC050 revision (the `RMII_CLK` naming difference).

## A test nothing runs is not a test

Two quiet ways for that to become true here: a tool grows a `--selftest` that the script never invokes, or the script stops being wired into the workflow. `tools/test_pipeline.sh` now asserts both rather than assuming them — it is the same class of mistake the selftests exist to catch.

Verified by breaking each:

```
$ # drop the check_hisi_padmux line from test_pipeline.sh
tools/check_hisi_padmux.py declares --selftest but tools/test_pipeline.sh never runs it
exit=1

$ # point the workflow step somewhere else
no CI workflow runs tools/test_pipeline.sh
exit=1
```

## Verified

Both selftests, `test_pipeline.sh`, native + `reginfo_test` + `cYAML_test`, arm32/mips32/arm64. No C changed.


widgetii:
All three fixed in `183eb02`. Every one was real, and they share a shape: **a guard that can pass without guarding anything.**

**1 — the linkage check protected itself.** Correct and the sharpest of the three. It lived inside `test_pipeline.sh`, so it only ran once the workflow had already run the script; delete that step and the check goes with it. Moved to the `unit-tests` job, which is required independently — dropping either job's step now leaves the other failing.

Writing the break-test for that turned up a **second hole in my own check**, which the review didn't catch either: it grepped the workflow for `tools/test_pipeline.sh`, a string its own line contains. It matched itself, so it could never fail. It now matches a `run:` line and nothing else:

```
### intact:                  exit=0
### pipeline step removed:   exit=1
```

**2 — discovery and wiring were both textual.** Agreed, and the recommended fix is better than what I wrote: how the option is spelled in the source is an implementation detail, and a grep for an invocation is equally satisfied by a comment. Both greps are gone. The script now asks each `tools/*.py` whether `--help` mentions `--selftest` and runs whatever answers — so a new tool needs no wiring, a renamed one cannot be silently dropped, and nothing counts as run unless it *ran*. An empty result is an error too, since losing the last one should be loud.

**3 — `--selftest` bypassed `parse_args`.** Real: `--selftest --typo` ran the fixture and exited 0. It is an ordinary flag now, with the operational arguments validated after parsing, which also fixes `--help`. `gen_sstar_padmux.py` had the identical bug from #196 and gets the identical fix — worth flagging since that one has been on master since then.

**Verified by breaking each:**

```
new tool with --selftest, no wiring   ->   selftest: tools/zz_probe.py   (picked up)
failing selftest                      ->   exit=1
no tool offers --selftest             ->   exit=1, "has one been lost?"
workflow step removed                 ->   exit=1
--selftest --bogus-option             ->   exit=2   (both tools)
```

Unchanged where it should be: EV20X, CV200, CV100 and AV100 still report `0 of N`, and `sstar_padmux.h` still regenerates byte-identical against the SDK.

PR #202, opened 2026-09-16
hisi: read the PINOUT workbooks, and check seven more tables

widgetii:
**I was wrong in #199.** I said the pin-mux register chapter stopped being shipped after Hi3516CV300 and that ~900 rows could not be checked. It didn't stop being shipped — it **moved out of the data sheet PDF and into the chip's `<chip>_PINOUT_*.xlsx`**:

| | where the register definitions live |
|---|---|
| up to Hi3518EV20X | data sheet PDF, chapter 2.3 |
| Hi3516CV300 | `<chip>_PINOUT_*.xlsx`, sheet `3.muxctrl_reg Description` |
| Hi3516CV500 onwards | `<chip>_PINOUT_*.xlsx`, sheet `3. Pin Control Registers` |
| Chinese editions | the same, as `3.管脚控制寄存器` |

I'd searched those SDK directories by filename only, in three of them, and never opened a workbook. The PDFs do still mention `iocfg_reg55` in passing and define none of them, which is what I generalised from — wrongly.

## Seven more tables check out

583 rows, all agreeing with their documents, **no change to `reginfo.c`**:

| table | SoC | rows | result |
|---|---|---|---|
| `CV300` | Hi3516C V300 | 66 | agrees |
| `EV200` | Hi3516E V200 | 52 | agrees |
| `EV300` | Hi3516E V300 | 93 | agrees |
| `_8EV300` | Hi3518E V300 | 56 | agrees |
| `DV200` | Hi3516D V200 | 99 | agrees |
| `CV500` | Hi3516C V500 | 104 | agrees |
| `DV300` | Hi3516D V300 | 113 | agrees |

Four of those ship only in Chinese, so the parse handles `功能选择：` with full-width punctuation and `0x`-prefixed values alongside the English `Function select` with decimal ones.

`--pinout` is the better of the two sources: it states **absolute addresses**, so there's no `--base` to get wrong and the multi-bank tables (`CV500`, `DV300`, `EV300`, `CV610`) need no special handling. `tools/xlsx_min.py` is thirty lines of `zipfile` + `ElementTree` rather than a dependency CI would have to install for one offline tool.

## Three parser gaps, each found as a table "disagreeing"

- **`MUXCTRL` rows end with `;` in the V5 tables** and without one everywhere else. Matching only the older shape found **1 of CV610's 76 rows** — which reads as a nearly empty table rather than as a parse that missed. (Only CV610 and DV500 are affected, both already unchecked, so no earlier result changes.)
- **`VOU1120_DATA2/ VOU656_DATA2`** is spaced in the Hi3516DV300 workbook and unspaced in the table. A spelling, not a rename — normalised. That was two of DV300's apparent disagreements.
- **`0xdeadbeef` is `IPCHW_PADMUX_ADDR_NONE`**, a pad the build deliberately cannot reach. No document has a register there and none should, so it is skipped rather than reported. That was CV300's only outlier.

The selftest grew cases for all three, plus the workbook field descriptions in both languages and both number bases. `classify()` now takes addresses rather than register numbers; `--base` is optional and is what still makes the alignment check possible on the PDF path. **All four PDF-checked tables still report `0 of N`.**

## Still open — checked this time, not assumed

- **`CV610`** (76 rows) genuinely disagrees with its workbook, and *not* as missing holes: functions absent, renamed, or merged into one slashed entry where the document gives two distinct values. The package matters — the **BGA** sheet disagrees on 65 rows, the **QFN** sheet and the **Hi3516CV608** sheet on 24 each — so which part the table was entered from has to be settled before anything is rewritten. Its own piece of work.
- **`AV200`** (Hi3519V101), **`DV500`** (Hi3519DV500), **`RCV100`** (Hi3536C), **`DV100`** (Hi3536D) — no document here in either form.

SDK source is no substitute for either: it shows the values a driver *writes*, and a hole is exactly the value nobody writes.

## Verified

`--selftest`, `test_pipeline.sh`, native + `reginfo_test` + `cYAML_test`, and every PDF table re-run. No C changed.


widgetii:
Fixed in `813c70a`.

**One correction to the premise first:** CV610's BGA, QFN and Hi3516CV608 pinouts are three separate **files**, selected with `--pinout <path>`, not three sheets of one workbook. So the specific scenario described can't occur. The recommendation stands on its own merits, though, and I've taken it.

`pinout_registers()` took the first sheet whose name matched. That is the failure that keeps recurring in this tool — the data sheets hold two SoCs' chapters and *that* path already refuses rather than guess, so the workbook path reading differently was an inconsistency waiting to bite. `pick_sheet()` now refuses when several match and names them, refuses when none does and lists what the workbook has, and takes `--sheet` to settle it.

**No workbook here has more than one matching sheet**, so this is insurance rather than a repair — I checked all 38 before deciding. But checking is what turned up a real gap:

> the **older Chinese** Hi3516CV300 and Hi3516EV100 workbooks call it `3.管脚复用寄存器` — a fourth spelling — and matched **nothing at all**.

They parse now. The near-miss next door is worth naming too: every one of these workbooks also carries a pin **drive capability** register list (`4.管脚驱动能力寄存器` / "Pin Drive Capability Registers"), which is not the mux and must not match.

That gained CV300 a second independent edition, and I then did the same for the rest. **All seven workbook tables now have two editions behind them** — an English and a Chinese workbook, or two SDK releases — agreeing on every selector value:

| table | rows | editions |
|---|---|---|
| `CV300` | 66 | EN + CN + SPC010 |
| `EV200` | 52 | SPC011 + SPC012 |
| `EV300` | 93 | SPC011 + SPC012 |
| `_8EV300` | 56 | SPC011 + SPC012 |
| `DV200` | 99 | SPC011 + SPC012 |
| `CV500` | 104 | EN + CN |
| `DV300` | 113 | EN + CN |

`pick_sheet()` is a pure function of the sheet-name list, so the selftest covers all four spellings, the drive-capability near-miss, the ambiguous case, `--sheet` overriding it, and `--sheet` naming something absent. PDF tables unchanged; `--selftest` and `test_pipeline.sh` green.

PR #205, opened 2026-09-16
hisi: AV200 checks out too, against the Hi3559/Hi3556 workbook

widgetii:
The last family listed as having no document. There is one — for two of the four SoCs `AV200` serves — and it was in the plain **Hi3559 V100** release, not either Hi3519 V101 release.

## Your hypothesis held up

`AV200` covers Hi3519 V101, Hi3559 V100, Hi3556 V100 and Hi3516A V200, and they do share a package:

| | Hi3519 V101 | Hi3516A V200 |
|---|---|---|
| package | FCCSP 10×10 mm, 0.4 mm pitch | identical |
| **total pins** | **524** | **524** |
| I/O / dig-pwr / dig-gnd / ana-pwr / ana-gnd | 268 / 61 / 161 / 14 / 20 | identical |

And the SDK's `.bsdl` and `.ibs` for the two are **byte-identical** (same SHA256); the file shipped as `Hi3516AV200.bsdl` declares `entity HI3519V101_TOP`.

## Result: 106 of 106 registers, 91 rows matching outright

Every register is in the document at the same address. The 15 that differ do so in **one direction only** — the table names an Ethernet function where the document has nothing or `reserved`:

```
AV200_muxctrl_reg80
  table     : GPIO10_1 VO_DATA18 RGMII_TX_CLK RMII_CLK
  data sheet: GPIO10_1 VO_DATA18
```

The full set of extras is `RGMII_TXD0-3`, `RGMII_RXD0-3`, `RGMII_TX_CLK/RX_CLK/TX_EN/RX_DV`, `RMII_CLK`, `MDIO`, `EPHY_CLK`, `EPHY_RSTN` — Ethernet, all of it.

That is exactly what the part names predict: **Hi3559/Hi3556 V100 are *HD Mobile Camera* SoCs** and don't bring Ethernet out; Hi3519 V101 and Hi3516A V200 are IP-camera parts that do, and are evidently what the table was entered from.

So **every function this document describes agrees** — no missing hole, no rename, no wrong address, exact register count. Both the English and Chinese editions give the same 15.

`Hi3519V101_PINOUT_CN.xls` would settle those fifteen. Its own user guide names it (*"请参见《HI3519V101_PINOUT_CN》.xls"*) and no release here ships it.

## One more sheet-naming allowance

Hi3559V100 writes `3.muxctrl_reg description` where Hi3516CV300 writes `3.muxctrl_reg Description`, so the match now folds case. None of the near-misses collide when folded, and `4.pad_ctrl_reg description` and `5.Hardware multiplexing` join the others in the selftest — which is what makes that claim checkable rather than asserted.

## Coverage after this

**Fourteen** tables verified against a document, each against two editions. Still genuinely open: **`CV610`** (package variant to settle first) and **`DV500`** (131 registers documented vs 102 rows, needs re-entering).

Verified: `--selftest`, `test_pipeline.sh`, and every other table re-run unchanged. No C changed.


widgetii:
Fixed in `84319dd`. The finding is right, and right in the way that matters most for that table.

The sharpest part is the one I'd missed entirely: **running the command the document tells you to run exits 1 for AV200.** So the coverage table asserted agreement that its own tool contradicts. That is worse than an overstatement — it is a claim a reader can falsify in ten seconds, which devalues every other row.

Three changes:

- the row now reads **`91 of 106`; the checker exits non-zero**
- the blanket sentence after the table now says "every row above **except `AV200`**… with the checker exiting 0"
- the row is explicitly to be read as **"no defect found", not "verified"**

I also went further than the finding asked, because the detail paragraph was overstated in a subtler way. It said *"every function this document does describe agrees: no missing hole, no rename, no wrong address"* — each clause true, and together they invite the wrong conclusion. The part-variant explanation (Hi3559/Hi3556 V100 are *HD Mobile Camera* parts with no Ethernet brought out, so a superset is expected) is a **reading, not a verification**:

> nothing here rules out a hole having been dropped inside those 15 rows the same way 19 were dropped from each V2 table, because the document that would show it describes a part that has no such pins

That is exactly the bug class this whole series exists to catch, and the Ethernet rows are the ones where it could still be hiding. The docs now say to treat Ethernet selectors on this family as unconfirmed.

Checked rather than asserted:

```
AV200 checker exit=1   (docs now say non-zero)
CV500 checker exit=0   (docs say 0)
```

widgetii:
Update in `8c1f26f` — **the Hi3519V101 workbook exists after all, and AV200 now verifies exactly.**

I said it was "referenced but not shipped". It is shipped, in the Hi3519 V101 **SPC040** release — and only there. **SPC050 is newer and carries no workbook at all**, and SPC050 is the release I checked. Checking the latest and concluding is the same mistake as searching by filename and concluding, one release along.

## Result

```
Hi3519 V101_PINOUT_EN.xlsx     0 of 106 rows disagree
Hi3519V101_PINOUT_CN.xlsx      0 of 106 rows disagree
Hi3516A V200_PINOUT_EN.xlsx    0 of 106 rows disagree
Hi3516AV200_PINOUT_CN.xlsx     0 of 106 rows disagree
```

And comparing the documents to each other rather than only to the table:

| | registers | verdict |
|---|---|---|
| Hi3519 V101 vs Hi3516A V200 | 106 each | **identical — 0 differ** |
| Hi3519 V101 vs Hi3559/Hi3556 V100 | 106 each | 15 differ, all Ethernet |

So the previous commit's reading is now **confirmed rather than assumed**: the parts really do differ only in Ethernet, and one table serving all four is right because the Ethernet-capable pair is a superset. The paragraph telling readers to treat Ethernet selectors as unconfirmed is gone, because they no longer are.

That also settles the question that started this: **Hi3519 V101 and Hi3516A V200 have the same pinout** — same package, same 524 pins, byte-identical BSDL/IBIS, and now identical mux maps register for register.

The `.xls` caveat goes too — the files are `.xlsx` whatever the user guide calls them.

The coverage row is back to `agrees`, this time meaning it. Worth noting the finding you raised still did its job: it is what stopped a row reading "agrees" on 91-of-106 evidence, and the row only says it now because the evidence changed.

PR #206, opened 2026-09-16
hisi: an incomplete pad-mux table is worth having; stop conflating it with a wrong one

widgetii:
The docs called `DV500` **"worse"** than `CV610` and said it *"needs re-entering"*. Both were the wrong way round for what a consumer actually cares about.

**`CV610` can give a wrong answer** — functions renamed, or merged into one slashed entry where the document gives two distinct values.
**`DV500` can only give less answer.** A pad a table doesn't describe is simply not offered — `by_pad()` returns nothing, `get()` says nothing, `set()` refuses — which is exactly how every pad behaves on a family with no table at all.

Incomplete degrades. Wrong misleads. Only one of those is a reason for caution.

## DV500 is incomplete in a way worth stating precisely

**91 of its 96 differing rows are a strict subset of the document** — every function it names sits at the right selector value. Nothing is shifted the way 19 rows of each V2 table were. What's missing is 29 registers with no row, and 44 rows carrying `"reserved"` at selector 0 where the document names the GPIO.

Those are gaps to fill a register at a time, additively. Not a defect to repair.

## The checker conflated them too

Before:

```
125 of 102 rows disagree (0 missing holes, 125 this tool will not touch)
```

Alarming, silent about which kind of gap, and counting past the number of rows. Now:

```
96 row(s) of 102 differ, and 29 register(s) have no row:
  0 missing hole(s), 91 incomplete but not wrong, 5 this tool will not touch
```

A row whose filled positions all match the document, and which is no longer than it, is classified `PARTIAL` and reported apart from the rest.

Exit status grows a third value — **0** agreeing, **2** when the only differences are incompleteness, **1** for anything unresolved. `DV500` exits 1 on the strength of its five remaining rows rather than its ninety-one gaps, which is the signal that's actually worth acting on.

## Checked

- exit **2** end to end, against a table trimmed to nothing but partial rows
- exit **1** for CV610 (24 unresolved) and DV500 (5 unresolved)
- exit **0** unchanged for EV20X, AV100, AV200, CV500, RCV100, DV200 …
- the V2 hole cases are still `HOLE`, not `PARTIAL` — a hole has a real function at a position the document calls `reserved`, which fails the subset test on the way past. That ordering is in the selftest, along with truncated rows, a blanked selector 0, a wrong name at a filled position, and a row longer than the document.

No C changed.


widgetii:
Fixed in `2c60a9f`. This one is a **regression this PR introduced**, not a pre-existing gap — worth saying plainly.

Reporting incompleteness separately meant moving absent registers and partial rows out of `unfixable`. The `--fix` write guard still tested only `unfixable == 0`, so a run with repairable holes **alongside** either kind of incompleteness would now write, where before it refused.

That is exactly the protection I argued for in #202, undone by my own refactor while I was looking at the reporting. And the failure it protects against is not hypothetical — it is the shape of this very series:

> a register the table has no row for, or a row filling only some of the document's selectors, is what pointing at a **related but wrong part** looks like — the Hi3559 workbook against the AV200 table is the worked example, a real document and a real family whose fifteen differing rows were simply not that chip

`--fix` now writes only when every difference is a missing hole, and names which kinds blocked it:

```
NOT rewriting anything: 130 difference(s) above are not missing holes
(0 this tool will not touch, 1 incomplete, 129 register(s) with no row).
```

**The second half of the finding was right too**, and I'd have missed it: `fixed` counts replacements made *in memory*, and the guard can decline to write them — so the old `fixed == holes` test could report success on a run that changed nothing. Success now requires the write to have actually happened.

Checked both directions rather than reasoned about:

```
one real hole + one incomplete row  ->  file byte-identical, exit 1
holes only                          ->  rewritten, exit 0 on the next run
```

Other exit codes unmoved — DV500 and CV610 exit 1, a partial-only table exits 2, the agreeing tables exit 0.

PR #207, opened 2026-09-16
hisi: CV610 is a family, and the table is right for one package of it

widgetii:
I had `CV610` down as *"genuinely wrong — functions absent, renamed, or merged into one slashed entry"*. **That was reading it against the wrong sheet.**

## Hi3516CV610 is five parts in two packages, and the packages encode differently

The data sheet says it outright:

> *"Hi3516CV610-10B/20S/20G uses the QFN package, while Hi3516CV610-00S/00G uses the TFBGA package."*

And `regs_by_chip()` hands `CV610regs` to everything under `HISI_OT` that isn't a DV500 — **Hi3516CV608, Hi3516CV610 and Hi3516CV613**.

| part | package | mux map | the table |
|---|---|---|---|
| Hi3516CV608 | QFN 9×9, only package | identical to CV610 QFN, register for register | **correct, 52/52** |
| Hi3516CV610-10B/-20S/-20G | QFN 9×9 | QFN map | **correct, 52/52** |
| Hi3516CV610-00S/-00G | TFBGA 12×13.3 | **differs on 25 of the 52 common registers** | wrong on those 25 |
| Hi3516CV613 | no document found anywhere | unknown | unverified |

The two packages genuinely differ in **encoding**, not just in which pads are bonded out:

```
0x1026002c  QFN:   GPIO3_6 SDIO0_CDATA1/EMMC_DATA1 reserved  reserved SPI0_CSN1 ...
            TFBGA: GPIO3_6 SDIO0_CDATA1            EMMC_DATA0 reserved SPI0_CSN1 ...
```

So the "merged slashed entry" I flagged as an error **is the QFN encoding, correctly transcribed**. The table matches the QFN sheet on all 52 registers, and the CV608 sheet is identical to it.

The 24 rows beyond that are TFBGA-only pads entered partially — no GPIO name, and `USB_OVRCUR` where that sheet says `USB2_OVRCUR`. On a QFN part they describe pads that aren't bonded out: noise, not a hazard.

## Nothing separates the packages at runtime

`CHIP_ID` at SC_CTRL `0x0EE0` — which `hisi_detect_cpu()` already reads — is `0x3516C610` for all five CV610 parts. `VENDOR_ID` at `0x0EEC` is `0x35`. The data sheet documents no other register in that block. `CHIP_ID` is an OTP readback, so a SKU *could* be fused there, but the document doesn't say it is and settling that needs boards of both packages.

Left as is deliberately: serving TFBGA would need a second table plus something to select it, and would otherwise break the package the table is currently right about.

## And the question that started this — kernel/U-Boot source

Checked for both CV610 and DV500, since it was the obvious next place to look. **Neither SDK carries the mux tables in source.** Both ship the kernel as a patch against vanilla 5.10 and declare the banks as devicetree syscon nodes for drivers to reach drive strength:

```
ioconfig0@10260000  /  ioconfig1@11130000     (CV610)
ioconfig0@0eff0000  /  ioconfig1@10260000     (DV500)
```

No function names, and two of DV500's four mux banks aren't referenced at all. My earlier reasoning — source shows what a driver *writes*, a hole is what nobody writes — holds, but it is now verified rather than assumed.

Docs only; no code change.


widgetii:
Two corrections in `776112d`, both from checking **where the id came from** rather than only reading the data sheet — and both in the direction of "less bad than I said".

**`CV613` is not a separate chip.** The commit that added it says so outright:

> `Add id 0x3516c613 for HI3516CV610 mod. 20S`

and `-20S` is one of the **QFN** parts. So `CV610regs` is *correct* for it — 52 of 52 — not "unverified" as my table claimed.

**That also reframes the TFBGA risk, downwards.** Two SKUs of the same silicon already report two different ids, so `CHIP_ID` is fused **per SKU**, not per die. Which means a `-00S`/`-00G` most likely reports an id of its own that this build doesn't know — in which case:

```
get_hisi_chip_id()  ->  "Got unexpected ID 0x..."  ->  "unknown"
chip_generation     ->  never set to HISI_OT
regs_by_chip()      ->  NULL
```

**No table rather than the wrong one** — the safe failure, and the one this code already prefers. The 25 wrong registers only land if a TFBGA part happens to report `0x3516C610` itself.

So the actionable item is narrower than I first wrote: what to capture from a TFBGA board is **its chip id**. If it collides with `0x3516C610` the fix is a second table selected on that id — not a change to `CV610regs`, which is right for the three QFN SKUs and the CV608.

Updated table:

| part | package | the table |
|---|---|---|
| Hi3516CV608 | QFN 9×9 | correct, 52/52 |
| Hi3516CV610-10B/-20S/-20G | QFN 9×9 | correct, 52/52 (`-20S` reports `0x3516C613`) |
| Hi3516CV610-00S/-00G | TFBGA 12×13.3 | wrong on 25 **if** it reports `0x3516C610`; otherwise no table |

PR #211, opened 2026-09-16
libipchw: compile only the HiSilicon generation the consumer runs on

widgetii:
Second step of the `ipcinfo` shrink series (after #210). This is the structural one: **22,116 bytes against the 35,124 a camera carries today**, for a gk7205v200-family image.

## The gap the existing knobs leave

`IPCHW_VENDORS` (#178) cannot reach the HiSilicon HAL at all, because `chipid.c` dispatches to it directly rather than through the vendor table. So `hal_hisi.c` has always been compiled in whole — a chip-ID table naming all ten generations, five sub-decoders, a sensor-bus back-end per generation, an eight-armed temperature reader, two die-ID readers. A consumer built against one vendor SDK runs on one generation and pays for the other nine.

`IPCHW_HISI` takes the shape and vocabulary of `IPCHW_PADMUX`: `all` (default), `none`, or a subset of `v1 v2 v2a v3 v3a v4 v4a v5 3536c 3536d`. The `ipctool` executable keeps every generation regardless, as with the other two knobs. `none` drops `hal_hisi.c` and `ispreg.c` outright, which is what a SigmaStar or Ingenic consumer wants.

## Measured

arm-openipc-linux-musleabi, in the firmware package's configure, from the 26,216 B `IPCHW_VENDORS=none` baseline:

| selection | bytes | |
|---|---:|---:|
| `HISI=all` | 26,216 | — |
| any single generation | 22,112–22,120 | ≈ −4,100 |
| `HISI=none` | 17,856 | −8,360 |

`VENDORS=none HISI=v4`, what a gk7205v200-family image wants: **22,116 B**, against **35,124 B** on a camera today.

There is a second, smaller cut in here: with every vendor HAL compiled out for the target architecture, `manufacturers[]` holds nothing but its sentinel, so `generic_detect_cpu()` cannot match and its three `/proc/cpuinfo` reads are pure cost. The count is a compile-time constant, so comparing against it folds and takes `line_from_file()` and the POSIX regex calls with it — `chipid.c.o` 2126 → 1531 B.

## Two things deliberately not gated

**The UART0-base arms are all-or-nothing**, on `IPCHW_HISI_ANY` rather than per generation. A base identifies a register layout, not a generation — `hisi_detect_cpu()` reads the chip ID to decide that — and there is **no arm for V4A at all**, so guarding them per generation silently left cv500/av300/dv300 with no path to detection. `IPCHW_HISI=v4a` measuring the same as `none` is what caught it.

**The four family IDs that set no `chip_generation`** (3520DV200, 3521V100, 3559AV100, the 0xBDA9D100 group) stay in every build. They are a name each, and dropping them turns a chip recognised today into `unknown`.

## Is the default still the default

At `IPCHW_VENDORS=all` / `IPCHW_HISI=all`, **every object is byte-identical to 13b73db except `hal_hisi.c.o`, which is 4 bytes smaller.** `hisi_ensure_sensor_enabled()`, `hisi_hal_cleanup()` and `setup_hal_hisi()` became switches, because an if/else chain interleaved with `#ifdef` leaves a dangling `else` when an arm is compiled out. The constants are mutually exclusive, so the transformation is equivalent, and a switch is what `hisi_get_temp()` and `hisi_get_die_id()` in these same files already use.

Keeping the rest byte-identical is why the HiSilicon sources are spliced back into their old position in the source list rather than appended — link order decides `.text` layout, and appending relaid the whole section for no reason.

## `ci: build every IPCHW_HISI selection, not just the default`

Each generation guards a different set of arms, so a guard that swallows a `#define` or leaves a dangling `else` breaks exactly one selection and no other. **Three did while this was being written** — `v5`, `v2;v2a` and `none` — and every one built fine as `all`.

The sweep builds `libipchw` alone, natively, for all thirteen selections; a few seconds. I confirmed it fails on the intended fault by putting a `#define` back inside a generation's `#ifdef`: `v5` failed, everything else passed.

## Verification

On a **hi3516ev200** (V4, SC2315E, musl armv7), three binaries side by side — master, `HISI=all`, `HISI=v4` — agree on `-c -f -v -s -l -F -i -S -x`, every long form, the combined `-ci`, and `-t`.

A deliberately wrong build (`HISI=v5` on that V4 board) fails honestly rather than guessing: `Got unexpected ID 0x3516e200 for HiSilicon` on stderr, `unknown` for the chip name, exit 1 for `-i`.

Twelve generation selections build clean, as do `IPCHW_VENDORS=all` and `=sstar` crossed with `IPCHW_HISI=none`, with no new warnings under `-Wextra -Werror=return-type`. `cYAML_test`, `reginfo_test` and `longse_test` pass, `reginfo_test` also when built with `IPCHW_HISI=v4`.


widgetii:
Thanks — two of the three are real and are fixed in c4c873a. The third I'm leaving, with reasoning.

**3. Vendor queries return blank output — correct, fixed.**

Real regression, and I'd missed it. `getchipvendor()` hands back `chip_manufacturer` whether detection succeeded or not, so the sentinel-only shortcut returning before filling the buffers made `ipcinfo -v` print an empty line. That matters beyond cosmetics: the firmware's `S70vendor` does `vendor=$(ipcinfo -v); load_"$vendor" -i`.

Reproduced natively, where the `manufacturers[]` table is sentinel-only by architecture:

```
master               -v -> [genuineintel]
branch before fix    -v -> []
branch after fix     -v -> [unknown]
```

Both buffers are now filled inside the branch, so the default build still folds the whole thing away. Note the shortcut still skips the `/proc/cpuinfo` read by design, so it reports `unknown` rather than the host CPU's `vendor_id` — which is the honest answer when no HAL could match, and not a path any camera build reaches (an `IPCHW_VENDORS=none` image is HiSilicon or Goke, and those detect through the UART0 base without consulting the table).

**2. One excluded chip bypasses filtering — correct, fixed.**

Also right, and it falsifies the reasoning I put in the commit message. I'd left `0xBDA9D100` unguarded on the grounds that the family IDs which set no `chip_generation` are worth keeping everywhere. That holds for three of its four sub-IDs — but `scsysid0 == 2` does set `HISI_3536C` and return `3536CV100`, so a build excluding 3536C recognised the chip and then had no temperature arm to service it. That sub-case is now guarded with the rest of its generation, and the reserved-value arm reports it instead.

**1. Unsupported boards pass chip detection — declining, and here is why.**

The behaviour is real, but it is master's, not something this PR introduces. `hisi_detect_cpu()` has always done:

```c
strcpy(chip_name, get_hisi_chip_id(family_id, SCSYSID[0] >> 24));
return true;
```

and `get_hisi_chip_id()`'s default arm has always printed `Got unexpected ID 0x%x for HiSilicon` and returned `"unknown"`. Any unrecognised HiSilicon silicon already yields a non-NULL `hiunknown` today. What this PR changes is the set of IDs that reach that arm, not what the arm does.

The suggested fix — return false for an unknown ID — would change behaviour for **every** board, including the ~90 that build the default configuration, and would trade a loud diagnostic plus `hiunknown` for silence plus a NULL. That is a defensible change to make, but it is a behavioural decision about unrecognised silicon, not a size one, and it does not belong in a PR whose safety argument is that the default build is byte-identical.

In practice the new route is also not a new failure mode: reaching it means running an image built for one generation on a board of another, where the kernel and vendor SDK are wrong anyway and the camera will not stream regardless of what `ipcinfo` prints. Verified on a V4 board with a deliberately mismatched `IPCHW_HISI=v5` build — `Got unexpected ID 0x3516e200 for HiSilicon` on stderr, `unknown` for the name, exit 1 for `-i`.

Happy to open a separate issue for making unrecognised HiSilicon IDs fail detection outright if that is wanted; it would want its own before/after on hardware.

**Re-verified after the fixes:** all thirteen `IPCHW_HISI` selections build clean with no new warnings; `cYAML_test`, `reginfo_test`, `longse_test` pass, `reginfo_test` also at `IPCHW_HISI=v4`; the default build is still byte-identical to 13b73db except `hal_hisi.c.o`'s four bytes; `VENDORS=none HISI=v4` is still 22116 bytes; and on a hi3516ev200 all 18 flags plus `-ci` still match a master-built binary.

PR #212, opened 2026-09-16
libipchw: let a consumer pick the sensor families it can drive

widgetii:
Third step of the `ipcinfo` shrink series (#210, #211). Unlike the first two, **this knob costs capability**, and that shapes both the default and the recommendation.

## What it does

`sensors.c` carries eleven hand-written probes and `get_sensor_id_i2c()` walks the lot in a fixed order. At `-Os` they are 5,388 bytes:

| | | | |
|---|---:|---|---:|
| sony | 1516 | soi | 248 |
| smartsens | 1364 | imagedesign | 184 |
| superpix | 576 | cvsens | 164 |
| omni | 572 | onsemi | 144 |
| galaxycore | 392 | techpoint | 132 |
| | | visemi | 96 |

`IPCHW_SENSORS` takes the shape of the knobs beside it — `all` (default), `none`, or a subset. The `ipctool` executable keeps every family regardless.

## The difference from the other two knobs

`IPCHW_VENDORS` and `IPCHW_HISI` drop code a given silicon **can never reach**. This one drops code it **might**. A camera wearing a sensor whose family was compiled out reports nothing at all, where a full build would still have named the part even when the image cannot drive it.

Demonstrated on a hi3516ev200 carrying an SC2315E:

```
smartsens kept     -s -> [sc2315e]   exit 0        (identical to master)
smartsens dropped  -s -> []          exit 1        -c still [hi3516ev200]
```

Downstream that becomes `SENSOR=unknown` and an abort in `load_hisilicon`. The `Config.in`-equivalent help text says as much, and the default stays `all`.

## Sizes, and why they are section totals

File size is a poor guide here — every selection quantises onto the same 4096-byte page boundaries, so `sony` and `sony;smartsens` can differ by 4,100 bytes of nothing. These are allocated section totals at `IPCHW_VENDORS=none` / `IPCHW_HISI=v4`, against **20,064** for all eleven:

| selection | content | delta |
|---|---:|---:|
| drop the five rarest | 19,072 | −992 |
| sony, smartsens, soi, omni, galaxycore | 18,421 | −1,643 |
| sony, smartsens | 16,946 | −3,118 |
| sony | 15,091 | −4,973 |
| none | 12,987 | −7,077 |

For a realistic HiSilicon image the honest figure is around **1 KB**. The larger numbers need a driver set narrow enough that dropping a family is genuinely safe. The bigger sensor win is the table-driven rewrite that comes next, which shrinks the probes without removing any.

## Mechanics

`get_sensor_id_i2c()`'s else chain became independent statements short-circuited on `detected` — probe order and first-match-wins unchanged, but an `#ifdef` between an `else` and its `if` leaves the chain dangling when a family goes. Sony takes the SPI path with it (`get_sensor_id_spi()` probes Sony and nothing else), and `dummy_change_addr` goes with that, because only that function installs it.

The unused `exit:` label in `get_sensor_id_i2c()` goes too — no `goto` has targeted it since it was written, and `-Wall` has always flagged it; this file is simply not built with `-Wall`. With it gone `sensors.c` has one warning left under `-Wall`, an unused parameter that predates this branch.

## Verification

**`IPCHW_VENDORS=all` and `=none`, both with `HISI=all` and `SENSORS=all`, are byte-identical to 7ba86ce.**

All 27 selections across the two sweeps build clean; the CI step from #211 now covers both knobs. `cYAML_test`, `reginfo_test`, `longse_test` pass, `reginfo_test` also at `HISI=v4` / `SENSORS=sony;smartsens`.

On the hi3516ev200, a build keeping sony, smartsens, soi, omni and galaxycore answers `-c -f -v -s -l -F -i -S -x`, every long form and `-ci` exactly as a master build does.


widgetii:
Closing: for OpenIPC this knob is the wrong trade.

A mainline image is per-SoC-family and gets flashed onto arbitrary retail cameras. `ipcinfo --short-sensor` runs exactly once, on an unprovisioned camera, and its answer is written to U-Boot env — so identifying a sensor nobody catalogued is the whole job. Compiling a probe family out turns that into `SENSOR is not detected, aborting` and a camera that never streams.

The size does not pay for that risk. The gk7205v200 driver set needs eight of the eleven families (os02g10 is found by the SuperPix probe, not OmniVision), leaving only techpoint, visemi and cvsens droppable: **536 bytes**.

The sensor win belongs in a table-driven rewrite of the probes, which shrinks them without removing any family.

Issue #157, opened 2026-05-04
IMX335 misdetected as IMX347 on register 0x3057 ambiguity


widgetii:
## Summary

`detect_sony_sensor` in `src/sensors.c:163-167` decides between IMX335 and IMX347 by reading register `0x3057`:

- `0x06` → IMX347 (returns immediately)
- `0x07` → IMX335 (handled later via `0x316A`)

In practice an IMX335 can read back `0x3057 = 0x06` transiently — observed after the chip has been driven through a sensor-stop / power-state cycle (in our case: a WDR mode session, then re-probe). The detector then commits to IMX347 and `try_sensor_config` fails (`No matches for 'imx347_i2c'`). A power-cycle of the sensor restores `0x3057 = 0x07` and detection succeeds.

This is the inverse of #28 (IMX347 → IMX335), which was closed by the same `0x06`/`0x07` switch we now hit from the other direction.

## Repro

- HiSilicon Hi3516EV300 + Sony IMX335 (MIPI 4-lane)
- Run `majestic` with WDR config (Mode=`WDR_MODE_2To1_LINE`, `[mipi] wdr_type = VC`), then kill it and restart in linear without power-cycling.
- Detection now returns `imx347` until the camera is rebooted.

Log on the failing run:

```
DEBUG <majestic> [sdk] try_sensor_config: Autodetected sensor as 'imx347_i2c'
ERROR <majestic> [sdk] find_sensor_config: No matches for 'imx347_i2c' in /etc/sensors directory
```

After power-cycle the very same binary sees `imx335_i2c` and starts cleanly.

## Suggested fix

Don't commit to IMX347 on `0x3057 == 0x06` alone. Cross-check against an IMX335-specific marker before deciding (the existing `0x316A & 0xFC == 0x7C` test in this same function would do, or `0x3078 == 0x01` per the HINT comment on line 185). If the IMX335 marker matches, prefer IMX335 even when `0x3057` momentarily reads `0x06`.

Reference: `src/sensors.c:156-189` in master.

PR #213, opened 2026-09-16
sensors: a behavioural fingerprint of the i2c probes, and three arms that failed it

widgetii:
Groundwork for the table-driven probe rewrite. That rewrite has one hard constraint — **sensor identification must not change** — and nothing in this repo can currently prove that, so this builds the proof first. It also catches three bugs on the way.

## Why not a table of expectations

The obvious test is a list of `{id, name}` pairs. It would be worthless here: it gets transcribed from the same source the refactor edits, so a transcription error copies straight across and the test cheerfully agrees with the bug. I wrote that extractor first and it happened to work — but it is the wrong shape.

## What this does instead

Sweeps **every value the ID registers can hold**, `0x0000`–`0xffff`, through the real probe against a mock bus, and records every one that yields a name. Nothing is read out of the source, so nothing can be mis-transcribed: `src/sensors_golden.txt` is what the code *does*. 1,125 parts across nine families, 17 KB.

A diff of that file is exactly the set of sensors whose identification changed. An empty diff says a rewrite named every sensor the old code named and named nothing new — the only claim worth making about work in here.

The probes are `static`, so the test includes `sensors.c` rather than linking it (and so must not also link `ipchw`). `hal/common.c` supplies the bus hooks the mock installs. `getchipvendor()`/`getchipname()` are stubbed, which keeps `chipid.c` and every vendor HAL out of the binary — and is also the only way to drive both sides of the two SmartSens IDs that name a different part depending on the SoC.

**Three things the sweep cannot reach are checked by hand.** Sony has no ID register — `detect_sony_sensor()` is a decision tree over initialisation defaults — so the two disambiguations field reports paid for are pinned directly: IMX335 vs IMX347 via the OB cropping defaults (#157), and IMX415 vs IMX335 via `0x3b00`/`0x300b`, including that a failed read must not pass for "not an IMX415" (#179). TechPoint whitelists nothing and turns any non-zero ID into `TP%04x`, so sweeping it would bury every other family under 65,535 near-identical lines; the rule is checked instead, zero included.

## Three arms that never named the part they found

The context handed to each probe is scrubbed with `0xa5` first, because `getsensorid()` hands them an uninitialised one — so a probe reporting success without writing a name shows up as `<uninitialised>` rather than as whatever the previous call left behind. That is how these surfaced, and the second commit's entire evidence is the three lines it moves:

```
-smartsens 2330 SC2330            +smartsens 2330 SC307H
-smartsens da23 <uninitialised>   +smartsens da23 SC1345
-smartsens dc42 <uninitialised>   +smartsens dc42 SC4336
```

`0x2330` writes `SC307H` then falls into the next arm, which breaks to the trailing `sprintf` and overwrites it with `"SC2330"` — the name is computed, discarded, and recomputed wrongly. `0xda23`/`0xdc42` correct the raw ID and then `return true` without reaching the `sprintf`, so they report success having written nothing, and the caller formats stack garbage. Every other correcting arm `break`s; `SC4336` and `SC1345` sit beside the existing `SC4336P` at `0x9c42` and `SC1346` at `0xda4d`.

None of the three is reachable on a lab board — nobody here has any of these parts — which is exactly why they survived.

## Verification

The harness earns its keep: two dumps are byte-identical, it passes against its own fingerprint, and changing one character of one sensor name (`SC450AI` → `SC451AI`) fails it with the id and both names.

`cYAML_test`, `reginfo_test`, `longse_test` and `sensors_test` all pass; the last is now wired into the unit-test job, so a regenerated fingerprint becomes a deliberate act with a diff to review rather than something noticed on a camera in a ceiling.


widgetii:
All four were right, and they were all the same hole: the sweep scripted each family's **primary** ID registers and nothing else, so every part reached through a second set of registers stayed outside the fingerprint. A test whose whole job is coverage cannot have that. Fixed in 182c079.

**The unit is now an identification path, not a family.** Four more, 32 parts:

| path | what it takes to get there | parts |
|---|---|---:|
| `omni-mfg` | wide read must say something the first switch does not know, then `0x301C/0x301D` must be `0x7f 0xa2`, then the product registers are read again a byte-address at a time | 8 |
| `galaxycore-legacy` | leaving the wide registers unanswered sends the probe to `0xf0/0xf1` — and that route reaches **both** its switches, not just the first | 16 |
| `superpix-mod` | `0x04` is the third byte of the wider signature, consulted only once the two-byte switch declines | 7 |
| `superpix-alt` | the second ID location at `0xfa/0xfb`, reached by answering the first with something non-zero that matches nothing | 1 |

`OS04D10` is the interesting one: it was **unreachable by construction**, not merely unscripted. With `0x04` unmapped the read returned `-1`, and `(msb << 16 | lsb << 8 | -1)` is `-1`, so the comparison against `0x530444` never had a chance to match. No amount of sweeping the two-byte space would have found it.

**On the widths** — also right, and it matters more than it looks. A real bus builds the address from `reg_width` and decodes the value by `data_width`, so the same register at a different shape is a different transaction. `detect_imagedesign_sensor` and `detect_visemi_sensor` both read `0x3000` and are told apart by data width **alone**. The mock now matches on all three.

Worth recording: **nothing already in the fingerprint moved.** All nine original paths produce exactly the same answers under the stricter mock, which is its own small result — it says the probes ask for what they appear to ask for.

Re-verified the harness still earns its keep, this time on one of the new paths: renaming `OS04D10` by one character fails it with the id and both names, where before this commit that sensor was invisible to the test.

Golden file: 1125 → 1157 lines.

